PR #1, opened 2024-07-25
Refactor, fix minor bugs, format/lint

Ash-L2L:


nchashch:
I like that you broke the validation rules up into separate functions. That is a good idea.

Eventually we should have functions implementing validation rules correspond to the sections in the specification text defining these rules. It would make sense to mention the actual function names in the specification as well.

PR #4, opened 2024-09-27
config: add RPC user+password, rework cookie auth

torkelrogstad:
1. Adds support for username + password when creating the Core RPC client
2. Removes the data dir parameter, and instead adds a parameter that takes in the cookie path directly. This removes the hardcoded 'regtest' network, as a byproduct.


octobocto:
Have tested this, works great!

PR #7, opened 2024-10-02
server: comments, error response for get_ctip, input validation

torkelrogstad:
1. Clarify what the `GetCtip` endpoint does through a comment in the proto file
2. Proper error handling in `GetCtip`: 
    1. Return an error if the sidechain is not found
    2. Make the sidechain number request field optional. This is to differentiate _not_ passing a number from _passing 0_. The former is invalid, and we indicate this with a proper error message
3. Update the README with more context for devving

### DIFF
--- a/proto/validator/v1/validator.proto
+++ b/proto/validator/v1/validator.proto
@@ -1,45 +1,41 @@
 syntax = "proto3";
 package validator.v1;
 
+import "google/protobuf/wrappers.proto";
+
 service ValidatorService {
   rpc ConnectBlock(ConnectBlockRequest) returns (ConnectBlockResponse);
   rpc DisconnectBlock(DisconnectBlockRequest) returns (DisconnectBlockResponse);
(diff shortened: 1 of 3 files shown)


proto/validator/v1/validator.proto
```diff
@@ -137,7 +131,10 @@ message Deposit {
 }
 
 message GetCtipRequest {
-  uint32 sidechain_number = 1;
+  // Sidechain number to fetch chain tip for. This argument is in fact
+  // NOT optional! It is marked as such to clearly differentiate it
+  // from passing in 0, which is a valid sidechain number.
+  optional uint32 sidechain_number = 1;
```
Ash-L2L, line 137: If this value is not optional, consider using `UInt32Value` https://protobuf.dev/reference/protobuf/google.protobuf/#uint32-value

Ash-L2L (changes requested):

proto/validator/v1/validator.proto
torkelrogstad: Fixed!

Ash-L2L (approved):
LGTM, approved

PR #8, opened 2024-10-02
Update README.md cargo run example

CryptAxe:
Is this an extra "cargo run --" ?


torkelrogstad (approved):
That's one too much `cargo run`! I don't have merge rights, though.

PR #9, opened 2024-10-04
multi: pull bip300301_wallet into enforcer

torkelrogstad:
1. Rename `Bip300` to `Validator`, to make it clearer what this does.
2. Copy over the wallet implementation from the `bip300301_wallet` repo.
3. Replaced usage of the `bs58` library with `rust-bitcoin`, since this is already in the dependency tree through BDK
4. Same goes for the `bip39` library
5. Note that we're using an outdated version of `cusf_sidechain_proto`. This is to minimize the changes done here.
6. Note that none of this is tested, nor exposed anywhere. This is also to minimize the changes needed in copying over the wallet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,15 @@ protox = "0.7.1"
 tonic-build = "0.12.3"
 
 [dependencies]
+bdk = { version = "0.29.0", features = ["all-keys", "sqlite"] }
 bincode = "1.3.3"
 bip300301_messages = { git = "https://github.com/LayerTwo-Labs/bip300301_messages" }
+bitcoin = "0.32.3"
+blake3 = "1.5.4"
 byteorder = "1.5.0"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -12,6 +12,8 @@ protox = "0.7.1"
 tonic-build = "0.12.3"
 
 [dependencies]
+bdk = { version = "0.29.0", features = ["all-keys", "sqlite"] }
+rusqlite = "0.28.0"
```
Ash-L2L, line 16: Please try to keep dependencies in lexicographic order.
We could use https://github.com/DevinR528/cargo-sort to enforce this, but I think it's not worth it for a single lint

Ash-L2L (changes requested):

Ash-L2L:
Requested minor changes
Otherwise, LGTM!

PR #10, opened 2024-10-04
ci: unify into single job

torkelrogstad:


.github/workflows/check_lint_build_release.yaml
```diff
@@ -27,189 +31,102 @@ jobs:
...
-
-      - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.5.1
-        with:
-          prefix-key: "ubuntu-stable"
-
-      - name: Rustfmt
-        run: cargo fmt --all -- --check
-      
       - name: Cargo check
```
Ash-L2L, line 42: This should happen after a cache restore

.github/workflows/check_lint_build_release.yaml
```diff
@@ -27,189 +31,102 @@ jobs:
...
       - uses: actions/checkout@v4
+
       - name: Install latest stable toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
+          components: rustfmt, clippy
+          target: ${{ matrix.name }}
 
       - name: Rust Cache
```
Ash-L2L, line 83: You need to use a cache key, so that the cache is not shared between targets

.github/workflows/check_lint_build_release.yaml
```diff
@@ -27,189 +31,102 @@ jobs:
...
+            name: x86_64-apple-darwin
+            binary-suffix: ""
+            rustflags: ""
+
+          - os: ubuntu-latest
+            name: x86_64-pc-windows-gnu
+            binary-suffix: .exe
+            rustflags: "-C linker=/usr/bin/x86_64-w64-mingw32-gcc"
+
+    name: Check, Lint, Build, Release (${{ matrix.name }})
```
Ash-L2L, line 68: This should be named eg. `Build, Release (${{ matrix.name }})`, if the check/lint step is skipped

.github/workflows/check_lint_build_release.yaml
```diff
@@ -27,189 +31,102 @@ jobs:
...
-    # env:
-    #  RUSTFLAGS: -D warnings
     steps:
       - uses: actions/checkout@v4
+
       - name: Install latest stable toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
+          components: rustfmt, clippy
```
Ash-L2L, line 80: These components are not needed if the `rustfmt` and `clippy` checks are not part of this job

Ash-L2L (changes requested):
* Checks/lints should  run before build/release stage
*  The Linux runner should maintain its Rust cache through checks/lints.
* Rust cache should not be shared between targets

.github/workflows/check_lint_build_release.yaml
```diff
@@ -6,17 +6,26 @@ env:
...
-      
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          components: rustfmt, clippy
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ matrix.name }}
```
Ash-L2L, line 22: `key` is an additional per-job specifier. If you want the cache to be stable across jobs, You need to set `shared-key` and perhaps `prefix-key`. See https://github.com/Swatinem/rust-cache/tree/master

`matrix.name` is not defined here. It needs to share the cache with `ubuntu-latest`/`x86_64-unknown-linux-gnu` in `build-release`.

.github/workflows/check_lint_build_release.yaml
```diff
@@ -27,189 +36,104 @@ jobs:
...
-        
       - name: Install mingw-w64
         run: sudo apt install mingw-w64
+        if: ${{ matrix.name == 'x86_64-pc-windows-gnu' }}
 
-      - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.5.1
-        with:
-          prefix-key: "x86_64-pc-windows-gnu"
+      - name: Rustfmt
```
Ash-L2L, line 97: This should be in `check-lint`

Ash-L2L (changes requested):

Ash-L2L (approved):
Caching across `check-lint` and `build-release` should be fixed in a follow-up PR.
Otherwise LGTM

PR #20, opened 2024-10-07
Implement Validator Service Server

Ash-L2L:


Cargo.toml
```diff
@@ -21,26 +22,35 @@ byteorder = "1.5.0"
 clap = { version = "4.5.18", features = ["derive"] }
 cusf_sidechain_types = { git = "https://github.com/LayerTwo-Labs/cusf_sidechain_types" }
 ed25519-dalek-bip32 = "0.3.0"
-env_logger = "0.11.5"
```
torkelrogstad, line 24: I'll admit I'm no Rust expert - but I really like env_logger. It seems like _a lot_ of Rust apps and libraries configure logging through `RUST_LOG`, notably that's how BDK does it. So by setting RUST_LOG it would be possible to our logs, as well as logs of our dependencies. Just my two cents!

src/main.rs
```diff
@@ -1,103 +1,102 @@
...
-        let mut wallet = Wallet::new(&cli, &validator)
+    let cli = cli::Config::parse();
+    set_tracing_subscriber(cli.log_level)?;
+
+    let mainchain_client = rpc_client::create_client(&cli.node_rpc_opts)?;
+    let (err_tx, err_rx) = futures::channel::oneshot::channel();
+    let validator = Validator::new(
+        mainchain_client.clone(),
+        cli.node_zmq_addr_sequence,
+        Path::new("./"),
```
torkelrogstad, line 69: should this be configurable through CLI opts?

src/main.rs
```diff
@@ -1,103 +1,102 @@
...
+        |err| async {
+            let _send_err: Result<(), _> = err_tx.send(err);
+        },
+    )
+    .into_diagnostic()?;
+
+    let wallet: Option<Wallet> = if cli.enable_wallet {
+        Some(
+            Wallet::new(
+                Path::new("./wallet-db"),
```
torkelrogstad, line 79: Same here, hard coded path seems bad

src/proto.rs
```diff
@@ -0,0 +1,484 @@
...
+        field_name: String,
+        message_name: String,
+    },
+    #[error("Unknown enum tag in field `{field_name}` of message `{message_name}`: `{tag}`")]
+    UnknownEnumTag {
+        field_name: String,
+        message_name: String,
+        tag: i32,
+    },
+}
```
torkelrogstad, line 36: Is this file some sort of generated code to do Proto message validation? Or is it written by hand?

src/main.rs
Ash-L2L: I think it should be, although this PR does not change the existing behavior. This can be made configurable in a future PR.

src/proto.rs
Ash-L2L: This is handwritten. I don't know of any codegen tool that can validate proto messages unfortunately.
Some of these error cases, eg `InvalidFieldValue` and `InvalidRepeatedValue`, could not be generated automatically, since they correspond to arbitrary validation logic

src/rpc_client.rs
```diff
@@ -0,0 +1,75 @@
...
+pub enum GetBlockchainInfoError {
+    #[error("No response data for rpc method `getblockchaininfo`")]
+    NoResponseData,
+    #[error("RPC error: `getblockchaininfo`")]
+    Rpc(#[from] Box<ureq_jsonrpc::Error>),
+}
+
+pub fn get_blockchain_info(client: &RpcClient) -> Result<BlockchainInfo, GetBlockchainInfoError> {
+    client
+        .send_request("getblockchain", &[])
```
torkelrogstad, line 72: It's `getblockchaininfo`, not `getblockchain`

src/main.rs
```diff
@@ -1,103 +1,102 @@
...
+    let tracing_subscriber = tracing_subscriber::registry()
+        .with(targets_filter)
+        .with(stdout_layer);
+    tracing::subscriber::set_global_default(tracing_subscriber)
+        .into_diagnostic()
+        .map_err(|err| miette::miette!("setting default subscriber failed: {err:#}"))
+}
+
+async fn run_server(validator: Validator, addr: SocketAddr) -> Result<()> {
+    tracing::info!("Listening for gRPC on {addr}");
```
torkelrogstad, line 41: This prints out `2024-10-07T17:16:38.121426Z  INFO bip300301_enforcer:41: Listening for gRPC on 127.0.0.1:50051` for me. Would it be possible to get it to print `src/main.rs:41` instead?

torkelrogstad (commented):
I'd suggest adding this to the README, under the Bitcoin Core requirement: 

1. Bitcoin Core, with ZMQ support. This needs be be running with the following
   flags:

```conf
user=user
password=password
signetblocktime=60
signetchallenge=00141f61d57873d70d28bd28b3c9f9d6bf818b5a0d6a

# this can also be set to a different address, as long
# as you also change the CLI parameter
zmqpubsequence=tcp://0.0.0.0:29000
```

Plus this further down, in the usage section: 

```
$ cargo run -- \
  --node-rpc-port=38332 \
  --node-rpc-user=user \
  --node-rpc-password=password \
  --node-zmq-addr-sequence=tcp://0.0.0.0:29000
```

src/rpc_client.rs
Ash-L2L: Nice catch, fixed!

src/main.rs
Ash-L2L: Done!

torkelrogstad:
I'd very strongly suggest that we change the [proto definitions](https://github.com/LayerTwo-Labs/cusf_sidechain_proto/blob/ash/proto/proto/cusf/validator/v1/validator.proto) where we take in `bytes`, and instead to `string` with hex-encodings. It makes it **a lot** easier to work bytes data from other places. An example: 

1. I'm trying to test the `GetBlockHeaderInfo` endpoint
2. I go [mempool.drivechain.live](https://mempool.drivechain.live/block/0000023bc2004ed6b1cf850e1ad9e0fd880432b4628762b40c13571dfa0c7ee1) and grab the most recent block hash: `0000023bc2004ed6b1cf850e1ad9e0fd880432b4628762b40c13571dfa0c7ee1`
3. I paste this into the request body with `buf curl`: `buf curl  --protocol grpc --http2-prior-knowledge -d '{"block_hash":"0000023bc2004ed6b1cf850e1ad9e0fd880432b4628762b40c13571dfa0c7ee1"}' http://localhost:50051/cusf.validator.v1.ValidatorService/GetBlockHeaderInfo`
4. I get an error, because I'm supposed to base64-encode the bytes...

src/validator/task.rs
```diff
@@ -0,0 +1,942 @@
...
+                    return Err(InitialSyncError::DecodeBlockHashHex {
+                        block_hash_hex: block_hash,
+                        source: err,
+                    })
+                }
+            }
+            .try_into()
+            .map(BlockHash::from_byte_array)
+            .unwrap()
+        };
```
torkelrogstad, line 847: This is incorrect. Block hashes must be deserialised in reverse! There's an implementation of this on the BlockHash type. 

```rust
let block_hash = {
            match BlockHash::from_str(&block_hash) {
                Ok(block_hash) => block_hash,
                Err(err) => {
                    return Err(InitialSyncError::DecodeBlockHashHex {
                        block_hash_hex: block_hash,
                        source: err,
                    })
                }
            }
        };
```

torkelrogstad: Which also means that `source` in `InitialSyncError.DecodeBlockHashHex` must be of type `HexToArrayError`

src/validator/task.rs
```diff
@@ -0,0 +1,942 @@
...
+        // validate block
+        // if invalid invalidate
+        // if valid connect
+        // wait 1 second
+        height += 1;
+        let block_hash = hex::decode(block_hash)
+            .into_diagnostic()?
+            .try_into()
+            .map(BlockHash::from_byte_array)
+            .unwrap();
```
torkelrogstad, line 914: Same comment here as above

src/validator/task.rs
```diff
@@ -0,0 +1,942 @@
...
+    }
+    tracing::debug!("Block height: {main_block_height}");
+
+    while height < main_block_height {
+        let block_hash: String = main_client
+            .send_request("getblockhash", &[json!(height)])
+            .into_diagnostic()?
+            .ok_or(miette!("failed to get block hash"))?;
+        let prev_blockhash =
+            bitcoin::consensus::encode::deserialize_hex::<BlockHash>(&block_hash).unwrap();
```
torkelrogstad, line 885: This is also incorrect, same byte endianness issue. 

```rust
let prev_blockhash = BlockHash::from_str(&block_hash).unwrap();
```

I guess the tip here was what you're referring to as sus in the comment below?

src/validator/task.rs
Ash-L2L: Nice catch, will fix this

src/validator/task.rs
Ash-L2L: The part I was referring to was 
```
 tracing::debug!("Mainchain tip: {prev_blockhash}");
 ```
 The mainchain tip should not be called `prev_blockhash`.

**Comment by Ash-L2L** [comment]
> I'd very strongly suggest that we change the [proto definitions](https://github.com/LayerTwo-Labs/cusf_sidechain_proto/blob/ash/proto/proto/cusf/validator/v1/validator.proto) where we take in `bytes`, and instead to `string` with hex-encodings.

I have mixed opinions on this. Hex encoding is certainly more readable. However it is likely to cause confusion,, since as you noted, Bitcoin displays block hashes with the bytes reversed.
The protobuf style guide also recommends against hex-encoding https://protobuf.dev/programming-guides/api/#dont-encode-data-in-a-string

If we must encode in hex strings, I would suggest using a wrapper that clearly indicates the encoding. IE.
```
message ConsensusHex {
  google.protobuf.StringValue hex = 1;
}
```

torkelrogstad:
On a general basis, I agree with the Protobuf style guide wrt to not hex-encoding bytes. Bitcoin-related hashes is more or less the only exception I have to this rule! I've been bitten by the irritating byte order in bitcoin hashes way too many times over multiple years and projects, and just dealing with hex-encoded strings has by far been the approach that has given me the least amount of head aches. IMO it's safe to assume that if there's a hex string, it's encoded in the way it would be displayed to a human. I.e. in a wallet, or on a block explorer

src/validator/task.rs
Ash-L2L: Fixed

src/validator/task.rs
Ash-L2L: Fixed

src/validator/task.rs
Ash-L2L: Fixed

torkelrogstad (approved):

PR #23, opened 2024-10-08
server+build: add gRPC reflection

torkelrogstad:


src/main.rs
```diff
@@ -39,8 +42,14 @@ fn set_tracing_subscriber(log_level: tracing::Level) -> miette::Result<()> {
 }
 
 async fn run_server(validator: Validator, addr: SocketAddr) -> Result<()> {
-    tracing::info!("Listening for gRPC on {addr}");
+    let reflection_service = tonic_reflection::server::Builder::configure()
```
Ash-L2L, line 45: Please set the service name. The file descriptor set contains two services, but the server only serves one of them. https://docs.rs/tonic-reflection/latest/tonic_reflection/server/struct.Builder.html#method.with_service_name

Ash-L2L (changes requested):

Ash-L2L:
LGTM, approved

PR #26, opened 2024-10-09
Use ZMQ streams instead of polling, split validator service

Ash-L2L:


Cargo.toml
```diff
@@ -44,9 +44,12 @@ tonic = "0.12.3"
 tonic-reflection = "0.12.3"
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
-ureq-jsonrpc = { git = "https://github.com/nchashch/ureq-jsonrpc" }
 zeromq = "0.4.0"
 
+[dependencies.bip300301]
+git = "https://github.com/Ash-L2L/bip300301.git"
+rev = "7406a92dd58c2b15c97d2bb4b1793946baf3bc2c"
+
```
torkelrogstad, line 52: What's going on here?

src/cli.rs
```diff
@@ -5,12 +5,13 @@ use std::{
...
+    SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 18443));
+
 #[derive(Clone, Args)]
 pub struct NodeRpcConfig {
-    #[arg(default_value = "localhost", long = "node-rpc-host")]
-    pub host: String,
-    #[arg(default_value = "18443", long = "node-rpc-port")]
-    pub port: u16,
+    #[arg(default_value_t = DEFAULT_NODE_RPC_ADDR, long = "node-rpc-addr")]
+    pub addr: SocketAddr,
```
torkelrogstad, line 14: This makes it impossible to specify `localhost` as the host. I think that's unfortunate. If we're going to consolidate the option (which makes sense!) this should be a string, or something that at least accepts `localhost:PORT`

src/proto.rs
```diff
@@ -97,6 +100,37 @@ pub mod validator {
...
+        pub fn decode<Message, T>(self, field_name: &str) -> Result<T, super::Error>
+        where
+            Message: prost::Name,
+            T: bitcoin::consensus::Decodable,
+        {
+            let ConsensusHex { hex } = self;
+            let hex = hex.ok_or_else(|| super::Error::missing_field::<ConsensusHex>("hex"))?;
+            bitcoin::consensus::encode::deserialize_hex(&hex)
+                .map_err(|_err| super::Error::invalid_field_value::<Message>(field_name, &hex))
+        }
```
torkelrogstad, line 113: This is not working for me. 

```
❮ buf curl  --emit-defaults --protocol grpc --http2-prior-knowledge -d '{"block_hash": {"hex": "00000221fa0e3aa42bf47a993be5d269551a249c1d33169c8be0ce5b098a385c"}}' http://localhost:50051/cusf.mainchain.v1.ValidatorService/GetBlockHeaderInfo
{
   "code": "unknown",
   "message": "Missing header info for block hash `5c388a095bcee08b9c16331d9c241a5569d2e53b997af42ba43a0efa21020000`"
}

**Review by torkelrogstad — COMMENTED** [review]
I'm able to do the initial sync, but I'm not able to fetch header info back

File: src/cli.rs (hunk 2 above)
Comment (torkelrogstad) [reply src/cli.rs:14]:
```
error: invalid value 'localhost:38332' for '--node-rpc-addr <ADDR>': invalid socket address syntax
```

This is rather unfortunate... But not a blocker, do as you want

**Review by torkelrogstad — APPROVED** [review]
Very nice!
```
❯ buf curl  --emit-defaults --protocol grpc --http2-prior-knowledge -d '{"block_hash": {"hex": "00000221fa0e3aa42bf47a993be5d269551a249c1d33169c8be0ce5b098a385c"}}' http://localhost:50051/cusf.mainchain.v1.ValidatorService/GetBlockHeaderInfo
{
  "headerInfo": {
    "blockHash": {
      "hex": "00000221fa0e3aa42bf47a993be5d269551a249c1d33169c8be0ce5b098a385c"
    },
    "prevBlockHash": {
      "hex": "000000fecf1d44c53f18d4e1e2fab3380e5c0280488141d3715368ae2866ec43"
    },
    "height": 14666,
    "work": {
      "hex": "14d4490000000000000000000000000000000000000000000000000000000000"
    }
  }
}
```

PR #28, opened 2024-10-09
validator: add function + test for creating sidechain proposal

torkelrogstad:
This partially implements the solution sketched out here: https://github.com/LayerTwo-Labs/cusf_sidechain_proto/pull/5#discussion_r1793646099

After https://github.com/LayerTwo-Labs/cusf_sidechain_proto/pull/5 is landed, the idea is to add an endpoint that exposes this functionality. 

~One issue: this is not behaving precisely as expected... I'm unsure if this is due do a bug in one of the upstream bip300 libraries, or due to me misunderstanding things. See https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/28/files#diff-a8412c4f5f378292029d100f76c5941c636af5cc429f504e82a636f1946fc3d2R1026-R1030~

The above was a misunderstanding on my part! Cleared up after talking with @Ash-L2L in private.

### DIFF
--- a/.github/workflows/check_lint_build_release.yaml
+++ b/.github/workflows/check_lint_build_release.yaml
@@ -1,6 +1,6 @@
 on: [pull_request, push]
 
-name: Check, Lint, Build, Release
+name: Check, Lint, Test, Build, Release
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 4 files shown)


Ash-L2L (approved):
LGTM, approved

PR #29, opened 2024-10-10
multi: add empty wallet gRPC service scaffold

torkelrogstad:
Some notes here: 
1. The gRPC endpoints are all unimplemented. I had hoped to get to actually implementing at least one, but setting up the scaffold was more work than I anticipated. 
2. I had to fight with the borrow checker **a lot**. Not because I do hard things here, but because I don't really know Rust. I might be doing things in a strange way! 
3. ~There's a complaint from Clippy that seems significant. No idea how to fix it... https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/29/files#diff-4db9c4cca2ff80c2ad80cbf7111959e148b31ee24cb354f8620c417047fe7b46R903-R908~

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -7,7 +7,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     io::Cursor,
     path::Path,
-    sync::{Arc, Mutex, MutexGuard},
+    sync::{Arc, Mutex},
     time::{SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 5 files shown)


src/wallet.rs
```diff
@@ -634,26 +664,37 @@ impl Wallet {
...
-                bdk::bitcoin::Transaction::consensus_decode(&mut cursor).into_diagnostic()?;
+                bdk::bitcoin::Transaction::consensus_decode(&mut Cursor::new(transaction_bytes))
+                    .into_diagnostic()?;
+
+            Some((ctip_outpoint, transaction))
+        } else {
+            None
+        };
+
+        let wallet = self.bitcoin_wallet.lock();
```
Ash-L2L, line 690: You could drop this lock early, instead of holding it to the end of the function, across several blocking db calls.
Consider enabling Clippy lints relevant to Mutex/RwLock guards -
https://rust-lang.github.io/rust-clippy/stable/index.html#/significant_drop_in_scrutinee
https://rust-lang.github.io/rust-clippy/stable/index.html#/significant_drop_tightening

src/wallet.rs
```diff
@@ -713,23 +758,22 @@ impl Wallet {
...
-                .into_diagnostic()?,
+                         WHERE sidechain_number = ?1;"
+            }
+            None => {
+                "SELECT sidechain_number, address, amount, tx_data
+                     FROM deposits INNER JOIN mempool ON deposits.txid = mempool.txid;"
+            }
         };
+
+        let connection = self.db_connection.lock();
```
Ash-L2L, line 773: Same here, this lock does not need to be held until the function returns

Ash-L2L (commented):
LGTM otherwise

PR #38, opened 2024-10-14
multi: implement create_bmm_critical_data_transaction

### LINKED ISSUE #43: Implement `create_bmm_critical_data_transaction` RPC

torkelrogstad:
Closes #43 

TODO: 
- [x] actually broadcast the TX
- [x] test this
- [x] Merge https://github.com/Ash-L2L/bip300301/pull/2 and update the dependency in `cargo.toml`

The code here has been used to construct seemingly valid BMM requests, for example here: https://mempool.drivechain.live/tx/48304f36b955f45901af77d2aa8f61893f7687ba3332601ad326da664bad1575

To try this for yourselves, you'll have to apply this patch in order to bypass the active sidechain + correct main hash checks:
```patch
diff --git a/src/server.rs b/src/server.rs
index 1d23b11..93f7802 100644
--- a/src/server.rs
+++ b/src/server.rs
@@ -574,6 +574,7 @@ impl WalletService for Arc<crate::wallet::Wallet> {
                 )
             })?;
 
+        /*
         match self.is_sidechain_active(sidechain_number).await {
             Ok(false) => {
                 return Err(tonic::Status::failed_precondition(
@@ -582,7 +583,7 @@ impl WalletService for Arc<crate::wallet::Wallet> {
             }
             Ok(true) => (),
             Err(err) => return Err(tonic::Status::from_error(err.into())),
-        }
+        } */
 
         // This is also called H*
         let critical_hash = critical_hash
@@ -603,7 +604,7 @@ impl WalletService for Arc<crate::wallet::Wallet> {
 
         // If the mainchain tip has progressed beyond this, the request is already
         // expired.
-        if mainchain_tip != bdk_block_hash_to_bitcoin_block_hash(prev_bytes) {
+        if false && mainchain_tip != bdk_block_hash_to_bitcoin_block_hash(prev_bytes) {
             let message = format!(
                 "invalid prev_bytes {}: expected {}",
                 prev_bytes, mainchain_tip

```

You can then try the endpoint: 

```bash
$ buf curl --protocol grpc --http2-prior-knowledge --emit-defaults -d '{"value_sats": 10000, "height": 123, "sidechain_id": 1, "critical_hash": {"hex": "1200007c0ca1efd8d128fedf50c73f395b0cceb0ffa823edbd971b4afd913021"}, "prev_bytes": {"hex": "000000e597b4ff2b7b81bcb2b0199b65471b9ece52b9688f18a8a7ed8e275eb1"}}' http://localhost:50051/cusf.mainchain.v1.WalletService/CreateBmmCriticalDataT
ransaction

{
  "txid": {
    "hex": "516a86d7e375cd3295288eaed9b691734e5a218e7e02c99fff6e8196977b84b4"
  }
}
```

Note that  in this case the hex printed to the console is _not_ the TXID of the broadcasted TX! This is because the `txid` field of the response here is of type `ConsensusHex`, and not `ReverseHex`. The actual TXID is `b4847b9796816eff9fc9027e8e215a4e7391b6d9ae8e289532cd75e3d7866a51`.


Ash-L2L:
The multiple versions of `bitcoin` is annoying, and something that happens quite frequently as the `bitcoin` crate often makes breaking API changes.
Trying to maintain a single version is futile. The best approach is to consistently use a recent version, and to use older versions compatible with dependencies as little as possible. Most crates using the `bitcoin` crate re-export it so that this is easy to do.

If it were reasonable to do so, I'd lint for this

```rust
use other_crate::bitcoin::{...};
```
except when aliased or anonymous, ie.
```rust
use other_crate::bitcoin::{
    self as other_crate_bitcoin, TxOut as OtherCrateTxOut, consensus::Encodable as _,
};
```

There are some exceptions though, eg. the wallet code needs to be compatible with BDK, so using `bdk::bitcoin` at the module scope is fine.


Since this occurs often, perhaps we should publish a crate that facilitates easy conversion between types from different versions of `bitcoin`?

src/server.rs
```diff
@@ -532,15 +530,114 @@ impl WalletService for Arc<crate::wallet::Wallet> {
...
+                invalid_field_value::<CreateBmmCriticalDataTransactionRequest>(
+                    "sidechain_id",
+                    &sidechain_id.unwrap_or_default().to_string(),
+                )
+            })?;
+
+        if !self
+            .is_sidechain_active(sidechain_number)
+            .await
+            .unwrap_or_default()
```
Ash-L2L, line 581: If the server encounters an internal error, it is probably best not to ignore that

src/server.rs
```diff
@@ -532,15 +530,114 @@ impl WalletService for Arc<crate::wallet::Wallet> {
...
+
+        let prev_bytes = prev_bytes
+            .ok_or_else(|| missing_field::<CreateBmmCriticalDataTransactionRequest>("prev_bytes"))?
+            .decode_tonic::<CreateBmmCriticalDataTransactionRequest, _>("prev_bytes")
+            .map(bdk::bitcoin::BlockHash::from_byte_array)?;
+
+        let mainchain_tip = self
+            .get_mainchain_tip()
+            .await
+            .map_err(|err| tonic::Status::internal(err.to_string()))?;
```
Ash-L2L, line 604: You should probably log an error here, if you are not including an error trace in the status message

src/server.rs
```diff
@@ -553,3 +650,23 @@ impl WalletService for Arc<crate::wallet::Wallet> {
         ))
     }
 }
+
+fn bdk_block_hash_to_bitcoin_block_hash(hash: bdk::bitcoin::BlockHash) -> bitcoin::BlockHash {
```
Ash-L2L, line 654: Nit: out-of-order declarations are very uncomfortable to read. Rust functions/blocks read top-to-bottom, so should modules unless there is a good reason not to

torkelrogstad:
> Since this occurs often, perhaps we should publish a crate that facilitates easy conversion between types from different versions of bitcoin?

Wouldn't a new create just make the problem even worse by introducing yet another version? Or perhaps it's possible to implement this in a version-agnostic way. If so, that'd be really nice.

src/wallet.rs
```diff
@@ -551,6 +557,92 @@ impl Wallet {
...
+
+        let psbt_finalized: Result<bool, _> = self
+            .bitcoin_wallet
+            .lock()
+            .sign(&mut psbt, SignOptions::default());
+        match psbt_finalized {
+            Ok(true) => {
+                tracing::info!("BMM request psbt signed successfully");
+            }
+
```
Ash-L2L, line 586: Why is there so much whitespace?

src/wallet.rs
```diff
@@ -514,7 +516,11 @@ impl Wallet {
         Ok(sidechains)
     }
 
-    pub async fn get_ctip(
+    pub async fn get_mainchain_tip(&self) -> Result<bitcoin::BlockHash> {
```
Ash-L2L, line 519: Is this actually async? I don't see an `await`

Ash-L2L (changes requested):
A few code quality/style issues/nits, but mostly LGTM.

nchashch:
This should close #43

nchashch:
> We're using two different version of rust-bitcoin throughout the project. One from bitcoin, and one re-exported from bdk. This leads to a bit wrangling back and forth to get the compiler to be happy. Would be very nice to converge on only one of these versions.

This should be fixed in a separate PR.

I opened an issue for this: #45

nchashch:
This RPC should insert a row into the `bmm_requests` table that was added in #41

src/errors.rs
```diff
@@ -0,0 +1,92 @@
...
+    // Add more here, as they appear
+    match err {
+        bdk::Error::Electrum(e) => convert_electrum_error(e),
+        _ => None,
+    }
+}
+
+fn convert_electrum_error(err: &bdk::electrum_client::Error) -> Option<miette::Report> {
+    match err {
+        bdk::electrum_client::Error::Protocol(err) => {
```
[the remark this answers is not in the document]
torkelrogstad: `err` is a JSON string that looks something like this: `electrum error: {\"code\": -10, \"message\": \"the error message\"}`. Hence the unescaping + dance back and forth... Updated the comments to make this clearer.

src/errors.rs
```diff
@@ -0,0 +1,92 @@
+use std::fmt;
+
+use miette::{diagnostic, miette, Diagnostic};
+use regex::Regex;
+use serde::Deserialize;
+
+pub fn convert_bdk_error(err: bdk::Error) -> miette::Report {
+    inner_convert_bdk_error(&err).unwrap_or_else(|| {
+        dump_error(&err);
```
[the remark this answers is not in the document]
torkelrogstad: The idea was to try and convert errors from `bdk` into something meaningful we can respond with in the API. And then somehow give us pointers to go on, if we're unable to construct a meaningful error code + message. Open to other ways of doing it!

src/server.rs
torkelrogstad: agreed!

src/wallet.rs
torkelrogstad: I guess this boils down to personal preference, but I find blocks of code much easier to parse when there's a little breathing room between them. Ideally the code formatter would handle absolutely everything, and there'd be no ambiguity in formatting. But whitespace is one of those things that formatters unfortunately aren't handling 100% strictly...

src/wallet.rs
torkelrogstad: You're right! Don't know how that got in there. Changing.

src/main.rs
```diff
@@ -10,6 +10,7 @@ use tower_http::trace::{DefaultOnFailure, DefaultOnResponse, TraceLayer};
 use tracing_subscriber::{filter as tracing_filter, layer::SubscriberExt};
 
 mod cli;
+mod errors;
```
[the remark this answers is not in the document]
torkelrogstad: I'll do that in a follow up PR that restructures the entire layout for the wallet functionality, if that's OK

src/server.rs
```diff
@@ -82,18 +83,16 @@ impl ValidatorService for Validator {
...
-                })?
-                .into()
-        };
+        let sidechain_id: SidechainNumber = sidechain_id
+            .ok_or_else(|| missing_field::<GetBlockInfoRequest>("sidechain_id"))
+            .map(SidechainNumber::try_from)?
+            .map_err(|_| {
+                invalid_field_value::<GetBlockInfoRequest>(
+                    "sidechain_id",
+                    &sidechain_id.unwrap_or_default().to_string(),
```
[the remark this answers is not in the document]
torkelrogstad: You're right, my bad! Fixing.

src/errors.rs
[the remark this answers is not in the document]
torkelrogstad: Added an attempt at a better approach in the last commit.

Ash-L2L (approved):

PR #53, opened 2024-10-21
cli: use sensible default for data_dir

octobocto:
By default, we set it to the platform specific datadir. .local/share for linux, Library/Application Support for mac, and APPDATA for windows


torkelrogstad (commented):
We now end up with this directory layout, which I think is unfortunate

```
/Users/torkel/Library/Application Support/bip300301_enforcer
├── bip300301_enforcer
│   └── signet.mdb
│       ├── data.mdb
│       └── lock.mdb
├── db.sqlite
└── wallet.sqlite
```

Could make sense to split things into `validator` and `wallet`, IMO

octobocto:
> We now end up with this directory layout, which I think is unfortunate
> 
> ```
> /Users/torkel/Library/Application Support/bip300301_enforcer
> ├── bip300301_enforcer
> │   └── signet.mdb
> │       ├── data.mdb
> │       └── lock.mdb
> ├── db.sqlite
> └── wallet.sqlite
> ```
> 
> Could make sense to split things into `validator` and `wallet`, IMO

Yes! That was a very confusing directory structure..

Ash-L2L (approved):
LGTM, approved

PR #54, opened 2024-10-21
validator: avoid erroring on invalid BMM requests

torkelrogstad:
Invalid BMM requests are not an error in our system, and should not stop our syncing progress.

Prior to this commit we would stop syncing at 26313 due to the inclusion of a BMM request that was not accepted by miners:

```
2024-10-21T06:26:22.262496Z ERROR bip300301_enforcer: src/main.rs:185: Failed to connect block: Error handling M8: Cannot include BMM request; not accepted by miners
```

### DIFF
--- a/src/validator/task.rs
+++ b/src/validator/task.rs
@@ -126,7 +126,7 @@ enum HandleM5M6Error {
 enum HandleM8Error {
     #[error("BMM request expired")]
     BmmRequestExpired,
-    #[error("Cannot include BMM request; not accepted by miners")]
+    #[error("BMM request not accepted by miners")]
     NotAcceptedByMiners,
 }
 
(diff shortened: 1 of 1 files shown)


src/validator/task.rs
```diff
@@ -796,16 +796,34 @@ fn connect_block(
...
+            Ok(true) => {
+                tracing::trace!(
+                    "Handled valid M8 BMM request in tx `{}`",
+                    transaction.compute_txid()
+                );
+            }
+            // This was not a BMM request
+            Ok(_) => (),
+
+            // Make sure to NOT propagate this outwards. These errors are not
```
Ash-L2L, line 813: If we want to distinguish between fatal and nonfatal errors, we should do so using something like https://docs.rs/fatality/latest/fatality/index.html

Ash-L2L (changes requested):

nchashch:
#55 seems to be better, because we are solving this problem systematically for all kinds of non-fatal errors, not just BMM errors. And it is better for readability, because you can look at the error enum variant and see if it is marked as `#[fatal]` -- with the approach in this PR you'd have to dig through the actual implementation code to see if an error is fatal or not.

nchashch:
I'll close this one, then we can decide between #55 and #56.

PR #55, opened 2024-10-21
Distinguish between fatal and nonfatal errors explicitly

Ash-L2L:
see https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/54

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.15"
+version = "0.6.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
+checksum = "23a1e53f0f5d86382dafe1cf314783b2044280f406e7e1506368220ad11b1338"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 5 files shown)


src/validator/task.rs
```diff
@@ -991,16 +1050,32 @@ pub(super) async fn task(
     zmq_addr_sequence: &str,
     dbs: &Dbs,
     event_tx: &Sender<Event>,
-) -> anyhow::Result<()> {
+) -> Result<(), Error> {
     // FIXME: use this instead of polling
```
torkelrogstad, line 1054: this comment looks outdated

src/validator/task.rs
```diff
@@ -88,79 +99,101 @@ enum HandleM3ProposeBundleError {
...
+    #[fatal(forward)]
     M3ProposeBundle(#[from] HandleM3ProposeBundleError),
     #[error("Error handling M4 (ack bundles)")]
+    #[fatal(forward)]
     M4AckBundles(#[from] HandleM4AckBundlesError),
     #[error("Error handling M5/M6")]
+    #[fatal(forward)]
     M5M6(#[from] HandleM5M6Error),
     #[error("Error handling M8")]
+    #[fatal(forward)]
```
torkelrogstad, line 196: It's a bit unfortunate that we have to mark errors as explicitly fatal. In a way I feel like it makes more sense to mark non-fatal errors as such? Seeing as those are going to be far less common

nchashch:
I'll merge this one, after the conflicts are resolved.

PR #56, opened 2024-10-21
validator: split validation issues from operation errors

torkelrogstad:
Alternative to #55 and #54 

Invalid MX messages are "okay" - in that they shoudln't crash our chain
sync progress. We therefore separate them from operational errors, such
as writing/reading to/from the database and other issues that IS in fact
things that should take us down.

Some of these message are just fine to ignore, whereas some of them
should result in blocl invalidation. This comes later!


nchashch:
I don't like that the `Ok` variant of the `Result` can contain an error, that can be confusing.

#55 is more idiomatic.

So I'll merge #55 and close this one.

PR #59, opened 2024-10-23
multi: mining improvements

torkelrogstad:
1. Give a clear error message that we cannot generate blocks on a non-regtest network
2. Add better error messages to the mining process

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -172,6 +172,18 @@ async fn main() -> Result<()> {
     }
 
     let mainchain_client = rpc_client::create_client(&cli.node_rpc_opts)?;
+
+    tracing::info!(
+        "Created mainchain client from options: {}:{}@{}",
+        cli.node_rpc_opts.user.as_deref().unwrap_or("cookie"),
+        cli.node_rpc_opts
+            .pass
+            .as_deref()
+            .map(|_| "*****")
+            .unwrap_or("cookie"),
+        cli.node_rpc_opts.addr,
+    );
+
     let (err_tx, err_rx) = futures::channel::oneshot::channel();
     let validator = Validator::new(
         mainchain_client.clone(),
--- a/src/wallet/error.rs
+++ b/src/wallet/error.rs
@@ -1,5 +1,6 @@
 use std::sync::LazyLock;
 
+use bip300301::jsonrpsee::core::client::Error as JsonRpcError;
 use miette::{diagnostic, Diagnostic};
 use regex::Regex;
 use serde::Deserialize;
@@ -103,3 +104,17 @@ pub fn convert_bdk_error(err: bdk::Error) -> miette::Report {
         Err(json_deserialize_err) => miette::Report::new(json_deserialize_err),
     }
 }
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("Bitcoin Core RPC error `{method}")]
+#[diagnostic(code(bitcoin_core_rpc_error))]
+pub struct BitcoinCoreRPC {
+    pub method: String,
+    #[source]
+    pub error: JsonRpcError,
+}
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("failed to consensus encode block")]
+#[diagnostic(code(encode_block_error))]
+pub struct EncodeBlock(#[from] pub bitcoin::io::Error);
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -220,7 +220,11 @@ impl Wallet {
             .main_client
             .get_blockchain_info()
             .await
-            .into_diagnostic()?;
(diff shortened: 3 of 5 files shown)


Ash-L2L (approved):
Left several style suggestions, but otherwise LGTM

nchashch:
LGTM, I'll merge this after the conflicts are resolved and the style suggestions are addressed.

nchashch (approved):

src/wallet/mod.rs
```diff
@@ -875,3 +884,17 @@ pub struct Deposit {
     pub amount: u64,
     pub transaction: Transaction,
 }
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("Bitcoin Core RPC error `{method}`: {error}")]
+#[diagnostic(code(bitcoin_core_rpc_error))]
+pub struct BitcoinCoreRPCError {
```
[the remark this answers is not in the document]
torkelrogstad: Moving these two errors into `wallet/error.rs`

src/main.rs
```diff
@@ -169,6 +169,12 @@ async fn main() -> Result<()> {
     }
 
     let mainchain_client = rpc_client::create_client(&cli.node_rpc_opts)?;
+
+    tracing::info!(
+        "Created mainchain client from options: {:?}",
```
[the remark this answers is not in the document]
torkelrogstad: Changed the log line to not displaying the entire struct. Just realized that'd be a bad idea anyways, considering there's a password field in there.

src/wallet/mod.rs
```diff
@@ -426,11 +436,7 @@ impl Wallet {
         with_connection(&self.db_connection.lock())
     }
 
-    pub async fn mine(
-        &self,
-        coinbase_outputs: &[TxOut],
-        transactions: Vec<Transaction>,
-    ) -> Result<()> {
+    async fn mine(&self, coinbase_outputs: &[TxOut], transactions: Vec<Transaction>) -> Result<()> {
```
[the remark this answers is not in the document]
torkelrogstad: I don't know if that is desirable, really. There's other error sources as well: 

1. SQLite errors from Rusqlite (get_bmm_requests)
2. Fetching BTC addresses from the wallet
3. SystemTimeError when fetching the current time
4. Probably more? These are the ones I've found so far

I could create a giant enum that covers all these cases (with accompanying error wrappers), but I don't really see the benefit, and think it'd end up being pretty cluttering.

src/wallet/error.rs
```diff
@@ -103,3 +104,21 @@ pub fn convert_bdk_error(err: bdk::Error) -> miette::Report {
         Err(json_deserialize_err) => miette::Report::new(json_deserialize_err),
     }
 }
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("SQLite error")]
+pub struct RusqliteError(#[from] pub rusqlite::Error);
```
Ash-L2L, line 110: Is this struct needed? You can just use `rusqlite::Error`

src/wallet/error.rs
```diff
@@ -103,3 +104,21 @@ pub fn convert_bdk_error(err: bdk::Error) -> miette::Report {
...
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("Bitcoin Core RPC error `{method}")]
+#[diagnostic(code(bitcoin_core_rpc_error))]
+pub struct BitcoinCoreRPCError {
```
Ash-L2L, line 115: Please remove the `Error` suffix from errors in an `error` module, so that they can be referred to as `error::BitcoinCoreRPC`, etc

src/wallet/error.rs
```diff
@@ -103,3 +104,21 @@ pub fn convert_bdk_error(err: bdk::Error) -> miette::Report {
...
+    pub method: String,
+    #[source]
+    pub error: JsonRpcError,
+}
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("failed to consensus encode block")]
+#[diagnostic(code(encode_block_error))]
+pub struct EncodeBlockError(#[from] pub bitcoin::io::Error);
```
Ash-L2L, line 124: Please remove the Error suffix from errors in an error module, so that they can be referred to as `error::EncodeBlock`, etc

src/wallet/mod.rs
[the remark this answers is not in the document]
Ash-L2L: OK, this can remain as-is

Ash-L2L (changes requested):

src/wallet/error.rs
torkelrogstad: Fixed

src/wallet/error.rs
torkelrogstad: Fixed

PR #69, opened 2024-10-30
multi: bump to BDK v1, electrum server adjustments

torkelrogstad:
Bump to BDK v1 (beta5). The idea was to bump to this so I could set up Bitcoin Core RPC as the chain backend on regtest, but this proved to be too complex to be worth it. The bump is nice, anyways!

Some other quality of life improvements: 

1. Verify the Electrum server is on the same network as we are on startup, giving a clear error message if this is not the case
2. Set dynamic default Electrum connection details, based on the network we're on. 
3. Add the current network to the data directory path. Without this, we get an error message when switching between regtest and signet. 
4. Update the README to include a section on the Electrum server part. 

I'm having some issues with the `significant_drop_tightening` Clippy lint. IMO we should drop it. It's a massive PITA with unclear benefits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,25 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
 }
 ```
 
+# Regtest
+
+By default, the enforcer runs against our custom signet. If you instead want to
+run against a local regtest, you need to also run a local regtest Electrum
+server.
+
+This can be done by:
+
+```bash
+$ git clone https://github.com/romanz/electrs
+
+# from within the cloned directory
+# note that electrs does not like user + password auth!
+# you'll have to used cookie-based authentication
+$ cargo run --release -- --network regtest \
+    --daemon-dir $HOME/.bitcoin \
+    --log-filters INFO
+```
+
 # Logging
 
 The application uses the `tracing` crate for logging. Logging is configured
--- a/README.md
+++ b/README.md
@@ -53,18 +53,26 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
 
 By default, the enforcer runs against our custom signet. If you instead want to
 run against a local regtest, you need to also run a local regtest Electrum
-server.
+server. There are multiple implementations of Electrum servers, an easy-to-use
+one is [`romanz/electrs`](https://github.com/romanz/electrs).
 
-This can be done by:
+For complete instructions on how to do this, consult the
+[official docs](https://github.com/romanz/electrs/blob/master/doc/install.md).
+
+A quickstart (that might not work, in case you're missing some dependencies):
 
 ```bash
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -49,6 +49,25 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
...
+run against a local regtest, you need to also run a local regtest Electrum
+server.
+
+This can be done by:
+
+```bash
+$ git clone https://github.com/romanz/electrs
+
+# from within the cloned directory
+# note that electrs does not like user + password auth!
```
Ash-L2L, line 64: `electrs` does support user+password auth via conf file.

`config.toml`
```
auth = "user:pass"
```

README.md
```diff
@@ -49,6 +49,25 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
...
+# you'll have to used cookie-based authentication
+$ cargo run --release -- --network regtest \
```
Ash-L2L, line 66: You need to change the working directory to `electrs` before doing this

README.md
```diff
@@ -49,6 +49,25 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
...
+    --daemon-dir $HOME/.bitcoin \
```
Ash-L2L, line 67: `daemon-dir` may be different, depending on how the user's node is configured. Best to just link to `electrs` docs

src/cli.rs
```diff
@@ -80,10 +80,15 @@ pub struct NodeRpcConfig {
 
 #[derive(Clone, Args)]
 pub struct WalletConfig {
-    #[arg(default_value = "drivechain.live", long = "wallet-electrum-host")]
-    pub electrum_host: String,
-    #[arg(default_value = "50001", long = "wallet-electrum-port")]
-    pub electrum_port: u16,
+    /// If no host is provided, a default value is used based on the network
+    /// we're on.
```
Ash-L2L, line 84: Can you document how the network is detected, and indicate the default values?

README.md
```diff
@@ -49,6 +49,25 @@ $ buf curl  --http2-prior-knowledge --protocol grpc \
...
+# Regtest
+
+By default, the enforcer runs against our custom signet. If you instead want to
+run against a local regtest, you need to also run a local regtest Electrum
+server.
+
+This can be done by:
+
+```bash
+$ git clone https://github.com/romanz/electrs
```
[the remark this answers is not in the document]
torkelrogstad: The electrs docs are a bit verbose. What do you think of linking to the docs, as well as a short-hand TLDR version?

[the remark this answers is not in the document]
torkelrogstad: Wrote this up in the README!

Ash-L2L (approved):

Issue #72, opened 2024-10-30
Wallet does not ensure that only one BMM accept is included per sidechain, per block


nchashch:
https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/master/src/wallet/mod.rs#L429

Issue #71, opened 2024-10-30
Wallet doesn't clear BMM requests after constructing a block


nchashch:
https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/master/src/wallet/mod.rs#L453

PR #74, opened 2024-10-31
Fix BMM ack issues

### LINKED ISSUE #70: M7 messages always ingored
M7 messages in a coinbase tx are always ignored, due to a missing case here
https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/master/src/messages.rs#L208

Ash-L2L:
Fixes #70, #71, #72

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -49,12 +49,35 @@ use crate::{
     cli::WalletConfig,
     convert,
     messages::{self, CoinbaseBuilder, M8_BMM_REQUEST_TAG},
+    proto::mainchain,
     types::{Ctip, SidechainAck, SidechainNumber, SidechainProposal},
     validator::Validator,
 };
 use error::WalletError;
 
 pub mod error;
+
+#[derive(Debug)]
+pub struct Deposit {
+    pub sidechain_number: u8,
+    pub address: Vec<u8>,
+    pub amount: u64,
+    pub transaction: Transaction,
+}
+
+fn get_block_value(height: u32, fees: Amount, network: Network) -> Amount {
+    let subsidy_sats = 50 * Amount::ONE_BTC.to_sat();
+    let subsidy_halving_interval = match network {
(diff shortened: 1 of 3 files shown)


torkelrogstad (approved):

src/wallet/mod.rs
```diff
@@ -49,13 +49,36 @@ use crate::{
...
     types::{Ctip, SidechainAck, SidechainNumber, SidechainProposal},
     validator::Validator,
 };
 use error::WalletError;
 
 pub mod error;
 
+#[derive(Debug)]
+pub struct Deposit {
+    pub sidechain_number: u8,
```
torkelrogstad, line 62: `SidechainNumber`?

src/wallet/mod.rs
```diff
@@ -49,13 +49,36 @@ use crate::{
...
+    pub address: Vec<u8>,
+    pub amount: u64,
```
torkelrogstad, line 64: `bitcoin::amount::Amount`?

torkelrogstad (approved):
Just minor stuff, not important. 

For next time: would be great if code moves were in separate commits from the actual changes. Makes it a lot easier to review the things that _do_ change

src/wallet/mod.rs
nchashch: I think u64 is better for internal stuff, `Amount` is useful for displaying and parsing Bitcoin amounts but there isn't much benefit to using it internally, it just adds more boilerplate.

Also it isn't the kind of situation where you are likely to make a type error -- you probably won't try to compute a sum of `amount` and `block_height`.

src/wallet/mod.rs
nchashch: I don't think the type wrapper for `sidechain_number` makes sense, it is already clear what it is just by the field name, adding the type wrapper just adds boilerplate without any clear benefit.

nchashch (approved):

PR #78, opened 2024-11-01
docs: add note on buf curl to README

### LINKED ISSUE #73: Implement a separate ergonomic CLI client to interact with the RPCs
That would improve development workflow.

torkelrogstad:
Closes #73


nchashch (approved):
Great! I like the detailed example.

PR #90, opened 2024-11-06
wallet: add balance and list transactions rpc

octobocto:
- **main: add debug logs to mainchain startup**
- **server+wallet: add get balance rpc**
- **rpc+wallet: add list transactions rpc**


torkelrogstad:
You can update the proto definitions in the sub module by: 

1. Committing changes in the sub module
2. Pushing to a branch on `cusf_sidechain_proto`
3. Updating the commit in the sub module by doing `git add cusf_sidechain_proto` from the root of this repo
4. Committing in this repo

There's also the option to mark a PR as draft, so that it gets properly marked as such in the UI

src/main.rs
```diff
@@ -178,7 +178,13 @@ async fn main() -> Result<()> {
     let info = mainchain_client
         .get_blockchain_info()
         .await
-        .into_diagnostic()?;
+        .map_err(|e| miette!("could not get blockchain info from mainchain client: {e}"))?;
```
torkelrogstad, line 181: it's better to use `error::BitcoinCoreRPC` here IMO

src/wallet/mod.rs
```diff
@@ -981,7 +981,8 @@ impl Wallet {
         tracing::trace!("Immature: {}", balance.immature);
         tracing::trace!("Untrusted pending: {}", balance.untrusted_pending);
         tracing::trace!("Trusted pending: {}", balance.trusted_pending);
```
torkelrogstad, line 983: Remove these?

Cargo.toml
```diff
@@ -41,6 +41,7 @@ nonempty = "0.10.0"
 parking_lot = { version = "0.12.3", features = ["send_guard"] }
 prost = "0.13.2"
 rand = "0.8.5"
+prost-types = "0.13.3"
```
torkelrogstad, line 44: what's this for?

src/server.rs
```diff
@@ -894,6 +898,44 @@ impl WalletService for Arc<crate::wallet::Wallet> {
         };
         Ok(tonic::Response::new(response))
     }
+
+    async fn list_transactions(
+        &self,
+        ListTransactionsResponserequest: tonic::Request<ListTransactionsRequest>,
```
torkelrogstad, line 904: that's a weird var name!

src/server.rs
```diff
@@ -894,6 +898,44 @@ impl WalletService for Arc<crate::wallet::Wallet> {
...
+    ) -> Result<tonic::Response<ListTransactionsResponse>, tonic::Status> {
+        let transactions = self
+            .list_wallet_transactions()
+            .await
+            .map_err(|err| err.into_status())?;
+
+        let response = ListTransactionsResponse {
+            transactions: transactions.into_iter().map(|tx| WalletTransaction {
+                txid: Some(ReverseHex::encode(&tx.txid.to_byte_array())),
```
torkelrogstad: Use `Some(ReverseHex::encode(&tx.txid))` instead

torkelrogstad: If there's a type mismatch from `bitcoin` vs. `bdk::bitcoin`, there's a `convert` module that either contains a function for converting between the types, or where you can add a function that converts

src/types.rs
```diff
@@ -321,6 +322,17 @@ pub enum Event {
...
 }
 
+#[derive(Debug)]
+pub struct WalletTransaction {
+    pub txid: bitcoin::Txid,
+    pub chain_position: ChainPosition<ConfirmationBlockTime>,
+    pub transaction: Arc<bitcoin::Transaction>,
+    pub fee_sats: i64,
+    pub received_sats: i64,
+    pub sent_sats: i64,
```
torkelrogstad, line 332: Probably best to use the `Amount` type here?

src/wallet/mod.rs
```diff
@@ -983,6 +983,59 @@ impl Wallet {
...
+        let mut txs = Vec::new();
+        
+        for tx in transactions {
+            // Calculate total input value
+            let mut input_value = 0;
+            for input in tx.tx_node.tx.input.iter() {
+                if let Ok(prev_output) = self.bitcoin_blockchain.fetch_tx(input.previous_output.txid) {
+                    input_value += prev_output.output[input.previous_output.vout as usize].value.to_sat();
+                }
+            }
```
torkelrogstad: This seems ... suboptimal. If we're fetching TXs from elsewhere we should instead use `getrawtransaction` from Bitcoin Core IMO, and avoid doing a RPC call _for each input_ in the TX. But do we really need to fetch anything from elsewhere?

src/wallet/mod.rs
```diff
@@ -983,6 +983,59 @@ impl Wallet {
         tracing::trace!("Trusted pending: {}", balance.trusted_pending);
         Ok(balance)
     }
+ 
+    pub async fn list_wallet_transactions(&self) -> Result<Vec<WalletTransaction>> {
+        let wallet = self.bitcoin_wallet.lock();
+        let transactions = wallet.transactions();
+        let mut txs = Vec::new();
+        
+        for tx in transactions {
```
torkelrogstad, line 992: Would be nice to run all of these in parallell. Not that I know precisely how to do that, off the top of my head

src/wallet/mod.rs
```diff
@@ -983,6 +983,59 @@ impl Wallet {
...
+            // Check outputs to wallet
+            for output in tx.tx_node.tx.output.iter() {
+                if wallet.is_mine(output.script_pubkey.clone()) {
+                    received_sats += output.value.to_sat();
+                }
+            }
+            
+            // Check inputs from wallet 
+            for input in tx.tx_node.tx.input.iter() {
+                if let Some(prev_output) = wallet.get_tx(input.previous_output.txid) {
```
torkelrogstad, line 1024: Same here as above, is this an I/O operation?

src/server.rs
```diff
@@ -29,22 +29,22 @@ use crate::{
         mainchain::{
             create_sidechain_proposal_response, get_bmm_h_star_commitment_response,
             get_ctip_response::Ctip, get_sidechain_proposals_response::SidechainProposal,
-            get_sidechains_response::SidechainInfo, server::ValidatorService,
-            wallet_service_server::WalletService, BroadcastWithdrawalBundleRequest,
+            get_sidechains_response::SidechainInfo, send_transaction_request::FeeRate,
```
torkelrogstad, line 32: This is in the cargo fmt commit: would be great if you could run cargo fmt on each commit instead of a separate formatting commit. Bad git hygiene! You can do a `git rebase -i origin/master` where you stop on each commit, run `cargo fmt` and then `commit --amend --no-edit`

src/wallet/mod.rs
octobocto: I don't know how else to do it. I don't get enough information from bdk.list_transactions. Only get general info of the prev output, but no amounts..

src/wallet/mod.rs
octobocto: The call is basically instant. I'm all for optimizing performance, but not before it's a problem!

src/types.rs
octobocto: Yes!

Cargo.toml
octobocto: It's for google proto timestamps

octobocto:
```protobuf.rs
pub struct Timestamp {
    /// Represents seconds of UTC time since Unix epoch
    /// 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to
    /// 9999-12-31T23:59:59Z inclusive.
    #[prost(int64, tag = "1")]
    pub seconds: i64,
    /// Non-negative fractions of a second at nanosecond resolution. Negative
    /// second values with fractions must still have non-negative nanos values
    /// that count forward in time. Must be from 0 to 999,999,999
    /// inclusive.
    #[prost(int32, tag = "2")]
    pub nanos: i32,
}
```

```server.rs line 932
                            timestamp: Some(Timestamp {
                                seconds: confirmation_time as i64,
                                nanos: 0,
                            }),
```

src/wallet/mod.rs
octobocto: Yes, this is an i/o operation! I'm currently running it in signet, and the call to listtransaction is as fast as CreateNewAddress

```
Executed in   39.73 millis    fish           external
   usr time   18.41 millis    0.09 millis   18.31 millis
   sys time   10.75 millis    1.08 millis    9.67 millis
```

src/wallet/mod.rs
Ash-L2L: It is possible to batch JSON-RPC requests, so you could batch-request `getrawtransaction`

torkelrogstad:
Needs rebase!

src/wallet/mod.rs
octobocto: I swapped to using getrawtransaction now

torkelrogstad (commented):
You should do the same trick you did with formatting with `cargo clippy`: `❯ cargo clippy --all-targets --all-features --fix --allow-dirty --allow-staged`

There's also some gnarly linter issues there that I'm guessing @Ash-L2L might need to help out with, a bit above our measly Rust pay grades

src/server.rs
```diff
@@ -904,6 +913,133 @@ impl WalletService for Arc<crate::wallet::Wallet> {
...
+                        ))
+                    })
+                    .map(|addr| (addr.assume_checked(), *amount))
+            })
+            .collect::<Result<HashMap<bdk_wallet::bitcoin::Address, u64>, tonic::Status>>()?;
+
+        let txid = self
+            .send_wallet_transaction(
+                destinations_validated,
+                fee_rate_per_vbyte,
```
Ash-L2L: If  `fee_rate_per_vbyte` and `fee` should not be set simultaneously, then this should be captured via the type system

Ash-L2L (changes requested):

Ash-L2L (approved):

PR #93, opened 2024-11-10
server/wallet: dont decode op_return_message

octobocto:
See https://github.com/LayerTwo-Labs/cusf_sidechain_proto/pull/28 for motivation and why this is necessary.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1002,12 +1002,6 @@ impl WalletService for Arc<crate::wallet::Wallet> {
             .transpose()
             .map_err(|err: crate::proto::Error| err.into_status())?;
 
-        let op_return_message = op_return_message
-            .map(|op_return_message| {
-                op_return_message.decode_tonic::<SendTransactionRequest, _>("op_return_message")
-            })
-            .transpose()?;
-
         let txid = self
             .send_wallet_transaction(destinations_validated, fee_policy, op_return_message)
             .await


torkelrogstad:
Don't think this is a good idea!

octobocto:
> Don't think this is a good idea!

How else can I pass raw bytes to the op-return output?

torkelrogstad:
I don't understand what you're saying here. The code as-is allows you to send raw bytes just fine. If you do this: 

```
$ buf curl  -d '{..., "op_return_message": {"hex": "deadbeef"}}' ... SendTransaction
```

That's going to end up as `Vec(0xde, 0xad, 0xbe, 0xef)` (i.e. raw bytes!), NOT as the string "deadbeef" in the final OP_RETURN message.

octobocto:
> I don't understand what you're saying here. The code as-is allows you to send raw bytes just fine. If you do this:
> 
> ```
> $ buf curl  -d '{..., "op_return_message": {"hex": "deadbeef"}}' ... SendTransaction
> ```
> 
> That's going to end up as `Vec(0xde, 0xad, 0xbe, 0xef)` (i.e. raw bytes!), NOT as the string "deadbeef" in the final OP_RETURN message.

I want to send `this is the embedded message`, and have that message be embedded. That does not work with the existing implementation.

torkelrogstad:
Then the correct thing to do is to decode that string into raw bytes, and then hex-encode those raw bytes. 

```go
hex.EncodeToString([]byte("this is the embedded message"))
``

Dunno the Rust equivalent off the top of my head, i'm sure chatgpt does

torkelrogstad:
Can this be closed?

PR #101, opened 2024-12-27
wallet: adjust miner script handling, add note on future GBT handling

torkelrogstad:


src/wallet/mod.rs
```diff
@@ -939,7 +939,13 @@ impl Wallet {
         command
             .current_dir(&dir)
             .arg("signet-miner/contrib/signet/miner")
+            .arg("--debug")
             .args(["--cli", "bitcoin-cli"])
+            // TODO: we can specify a custom command to invoke to specifically
```
Ash-L2L, line 944: Should this be included in this PR?

src/wallet/mod.rs
```diff
@@ -949,6 +955,13 @@ impl Wallet {
         tracing::info!("Running signet miner: {:?}", command);
 
         let output = command
+            // Send the stdout directly to our own stdout.
```
Ash-L2L, line 958: The signet miner script logs on stderr, so it would be a good idea to capture both.

torkelrogstad:
This was superseded by later PRs, so I'm actually going to close this

PR #104, opened 2024-12-30
multi: implement wallet initialization from BIP39 seed

torkelrogstad:
The wallet is now no longer initialized with a hard coded BIP39 seed. In this PR we add support for both plaintext + encrypted mnemonics. They can either be passed in through the new `CreateWallet` endpoint, or generated automatically. 

For testing/dev purposes we also add a new `--wallet-auto-init` flag that generates a brand new, plaintext mnemonic if it doesn't already exist. 

TODO: 
- [x] fix existing integration tests
- [ ] add more integration tests
- [ ] document behavior in the README
- [ ] probably flesh out some of the CLI doc comments

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,13 +12,12 @@ protox = "0.7.1"
 tonic-build = "0.12.3"
 
 [dependencies]
+aes-gcm = "0.10.3"
 anyhow = "1.0.89"
+argon2 = "0.5.3"
 async-broadcast = "0.7.1"
 bdk_electrum = "0.20.1"
-bdk_wallet = { version = "1.0.0", features = [
-    "file_store",
-    "keys-bip39",
-] }
+bdk_wallet = { version = "1.0.0", features = ["file_store", "keys-bip39"] }
 bincode = "1.3.3"
 bitcoin = "0.32.3"
 blake3 = "1.5.4"
@@ -41,8 +40,8 @@ nonempty = "0.11.0"
 ordermap = { version = "0.5.3", features = ["serde"] }
 parking_lot = { version = "0.12.3", features = ["send_guard"] }
 prost = "0.13.2"
-rand = "0.8.5"
 prost-types = "0.13.3"
+rand = "0.8.5"
 regex = "1.11.0"
 rusqlite = { version = "0.32.1", features = ["bundled"] }
 rusqlite_migration = "1.3.1"
@@ -59,8 +58,6 @@ tower-http = { version = "0.6.1", features = ["trace"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 zeromq = "0.4.1"
-argon2 = "0.5.3"
-aes-gcm = "0.10.3"
 
 [dependencies.bip300301]
 git = "https://github.com/Ash-L2L/bip300301.git"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,8 @@ tower-http = { version = "0.6.1", features = ["trace"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 zeromq = "0.4.1"
+argon2 = "0.5.3"
+aes-gcm = "0.10.3"
 
 [dependencies.bip300301]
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -59,6 +59,8 @@ tower-http = { version = "0.6.1", features = ["trace"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 zeromq = "0.4.1"
+argon2 = "0.5.3"
```
Ash-L2L, line 62: Please keep dependencies in lexicographic order where possible

src/cli.rs
```diff
@@ -93,6 +93,11 @@ pub struct WalletConfig {
     /// Signet: 50001, regtest: 60401
     #[arg(long = "wallet-electrum-port")]
     pub electrum_port: Option<u16>,
+
+    /// If no existing wallet is found, automatically initialize a new one.
+    /// Creates an unencrypted, plaintext wallet.
+    #[arg(long = "wallet-auto-init", default_value_t = false)]
```
Ash-L2L, line 99: I think `initialize` is too indirect, it should be explicit that a *new* wallet will be created from a *new* seed.
Perhaps the arg name should change too

src/main.rs
```diff
@@ -197,6 +197,7 @@ async fn mempool_task<Enforcer, RpcClient, F, Fut>(
             Ok(res) => res,
             Err(err) => {
                 let err = anyhow::Error::from(err);
+                tracing::error!("mempool sync error: {err:#}");
```
Ash-L2L, line 200: This error will be logged by the error handler - if you want to make it explicit that it is a mempool sync error, either do
```
let err = anyhow::anyhow!("mempool sync error: {err:#}")
```

or use a wrapper error struct that has this message wrapper, around an inner enum for the error cases

src/wallet/mod.rs
```diff
@@ -56,69 +59,56 @@ use crate::{
...
 
+type WrappedBdkWallet = Arc<Mutex<bdk_wallet::PersistedWallet<file_store::Store<ChangeSet>>>>;
+
 struct WalletInner {
     main_client: HttpClient,
     validator: Validator,
-    bitcoin_wallet: Mutex<bdk_wallet::PersistedWallet<file_store::Store<ChangeSet>>>,
+    // Unlocked, ready-to-go wallet: Some
+    // Locked wallet: None
+    bitcoin_wallet: RwLock<Option<WrappedBdkWallet>>,
```
Ash-L2L, line 73: Why is there an RwLock over a type that contains an `Arc<Mutex<...>>`?
This looks like it should be
`Option<Mutex<bdk_wallet::PersistedWallet<file_store::Store<ChangeSet>>>>`

Ash-L2L (changes requested):

src/wallet/mod.rs
torkelrogstad: I originally tried that (+ other variations over the same concept), and this was the only way I was able to get the type checker happy. The crux is in the parts where we assign to `self.bitcoin_wallet`, I couldn't get that to work without a `RwLock` + `Arc<Mutex<...>>`. If you're able to simplify the types, I'm very much open to that!

src/cli.rs
torkelrogstad: good point! Renamed to `wallet-auto-create`

src/wallet/mod.rs
Ash-L2L: Fixed in c948cd5459fee1a3e8a05be4c83dc17b9f65bc4c

integration_tests/util.rs
```diff
@@ -435,26 +454,25 @@ pub struct SignetMiner {
 }
 
 impl SignetMiner {
-    pub fn command<CmdArg, Subcommand, SubcommandArg, CmdArgs, SubcommandArgs>(
```
Ash-L2L, line 438: Why restrict this to Vec/&str? This is less flexible than before, which handled Vecs/arrays/ &str/String

Ash-L2L: Otherwise, LGTM. I'll merge this, can address this comment in a follow-up PR if necessary

PR #106, opened 2025-01-01
multi: parameterize more signet mining components

torkelrogstad:
1. implement signet mining within the wallet in terms of the integration
   test util harnesses
2. parameterize more input components: `bitcoin-cli` path,
   `bitcoin-util` path and signet mining script path


integration_tests/util.rs
```diff
@@ -442,53 +366,21 @@ impl Enforcer {
...
-                command.arg(format!(
-                    "--getblocktemplate-command={getblocktemplate_command}"
-                ));
-                if self.coinbasetxn {
-                    command.arg("--coinbasetxn");
-                }
-            }
+/// Cleans up a temporary directory, but without fatally exiting on errors that
+/// aren't that bad.
+pub fn drop_temp_dir(dir: &TempDir) -> Result<(), std::io::Error> {
```
Ash-L2L, line 371: This does not actually drop the `TempDir`, should be renamed

integration_tests/util.rs
```diff
@@ -442,53 +366,21 @@ impl Enforcer {
...
+    match dir.cleanup() {
+        Ok(_) => Ok(()),
+        // Seeing intermittent errors on CI related to deleting non-empty
```
Ash-L2L, line 374: This error occurs because a write to a temp dir occurs while deleting it. The fix is to kill running processes before deleting the temp dir.

src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct NodeRpcConfig {
     pub pass: Option<String>,
 }
 
+#[derive(Clone, Args)]
+pub struct MiningConfig {
+    /// Path to the Python mining script from Bitcoin Core. If not set,
+    /// the mining script is downloaded from GitHub.
+    #[arg(long = "signet-miner-script-path")]
+    pub signet_script_path: Option<PathBuf>,
```
Ash-L2L, line 86: This field name should be improved, `signet_miner_script_path` is good

src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct NodeRpcConfig {
...
+
+    /// Path to the Bitcoin Core `bitcoin-util` binary. Defaults to `bitcoin-util`.
+    #[arg(long = "signet-miner-bitcoin-util-path")]
+    pub signet_bitcoin_util_path: Option<PathBuf>,
```
Ash-L2L, line 90: is `signet` needed in the field name here?

src/cli.rs
```diff
@@ -78,6 +78,22 @@ pub struct NodeRpcConfig {
...
+
+    /// Path to the Bitcoin Core `bitcoin-cli` binary. Defaults to `bitcoin-cli`.
+    #[arg(long = "signet-miner-bitcoin-cli-path")]
+    pub signet_bitcoin_cli_path: Option<PathBuf>,
```
Ash-L2L, line 94: is `signet` needed in the field names here?

src/wallet/mine.rs
```diff
@@ -464,29 +487,38 @@ impl Wallet {
     // from Bitcoin Core. We assume that validation of this request has
     // happened elsewhere (i.e. that we're on signet, and have signing
     // capabilities).
-    async fn generate_signet_block(&self, address: &str) -> miette::Result<BlockHash> {
+    async fn generate_signet_block(&self) -> miette::Result<BlockHash> {
```
Ash-L2L, line 490: This function is becoming very large, can it be refactored?

src/wallet/signet_miner.rs
```diff
@@ -0,0 +1,150 @@
+use std::{ffi::OsStr, future::Future, path::PathBuf, time::Duration};
+
+/// Run command with args, dumping stderr/stdout to `dir`` on exit
+pub fn await_command_with_args<Cmd, Env, Arg, Envs, Args>(
```
Ash-L2L, line 4: This is not the right place for this function (should be in some kind of `util` module, since it is public and used in tests)
Also inappropriate to dump stdout/stderr in this manner from an application. This is fine in tests that dump to temporary directories, but not in an application.

This should go back to `integration_tests::util`, and `SignetMiner::command` should use `tokio::process::Command` directly in the implementation

src/wallet/signet_miner.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct SignetMiner {
+    /// Path to the Python mining script from Bitcoin Core.
+    pub path: PathBuf,
+
+    /// Command to use for executing `bitcoin-cli`. If not set,
+    /// we'll simply use `bitcoin-cli`.
+    pub bitcoin_cli: Option<String>,
```
Ash-L2L, line 89: Why change this to `String` when the `BitcoinCli` struct already constructs the `--cli=` arg correctly?

Ash-L2L (changes requested):

Ash-L2L (changes requested):
IMO, it would be best to define a module that exposes `BitcoinCli`, `Bitcointil`, and `SignetMiner` structs. These structs should make it possible to define how stderr/stdout should be handled

src/cli.rs
torkelrogstad: The reasoning from my side is that this is only needed when mining on signet, and therefore I wanted to indicate that this was a signet-related option. I tried to find functionality in Clap for grouping arguments, but I couldn't figure out how to do that...

src/cli.rs
torkelrogstad: Same reasoning as here https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/106#discussion_r1900912271

src/wallet/signet_miner.rs
torkelrogstad: Because I wanted to simplify how I passed in the argument here: https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/6bc7038a4aa54c7c883245c646139754127c39f9/src/wallet/mine.rs#L552-L559

I guess one alternative would be to make the fields on BitcoinCli `Option`s, and dynamically adjust which flags were passed based on which option was set

integration_tests/util.rs
torkelrogstad: That sounds like a better approach. I tried to return the `tokio::process::Command` from `await_command_with_args`, but quickly ran into issues with the borrow checker. The idea was to return the command itself, so that it could be killed when we needed to. Is there a better way to do it? And perhaps you'll have more luck fighting the compiler?

torkelrogstad:
Closing in favor of #122

PR #109, opened 2025-01-02
Refactor integration tests

Ash-L2L:


integration_tests/util.rs
```diff
@@ -170,6 +187,34 @@ where
...
+    Env: AsRef<OsStr>,
+    Envs: IntoIterator<Item = (Env, Env)>,
+    Args: IntoIterator<Item = Arg>,
+    F: FnOnce(anyhow::Error) + Send + 'static,
+{
+    let fut = await_command_with_args(dir, command, envs, args);
+    tokio::task::spawn(async {
+        use tracing::Instrument as _;
+        let err = fut.in_current_span().await;
+        tracing::error!("Command failed with error: {err:#}");
```
torkelrogstad, line 212: Shouldn't this log be emitted only if there's actually an error?

torkelrogstad (approved):
As a preparatory step in writing tests for the wallet seed handling logic I actually had something going locally which is very similar to this in spirit

integration_tests/util.rs
Ash-L2L: The log is only emitted if there is actually an error.
Note that `await_command_with_args`j returns an error because it is expected to run indefinitely otherwise.
```rust
await_command_with_args: Fn<Cmd, Env, Arg, Envs, Args>(
    &std::path::Path,
    Cmd,
    Envs,
    Args,
) -> impl Future<Output = anyhow::Error> + 'static
```

and so here, `err: anyhow::Error`.

PR #122, opened 2025-01-13
multi: parameterize more signet mining components

torkelrogstad:
Supersedes #106. I had issues with rebasing, so I ended up saying "screw this" and just reimplement the same thing on master by copying over code without the git history.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1,7 +1,6 @@
 use std::{
     borrow::BorrowMut,
     collections::HashMap,
-    net::SocketAddr,
     path::Path,
     str::FromStr,
     sync::Arc,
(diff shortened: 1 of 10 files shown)


src/wallet/mod.rs
```diff
@@ -626,7 +625,7 @@ impl Drop for Task {
 /// Cheap to clone, since it uses Arc internally
 #[derive(Clone)]
 pub struct Wallet {
-    serve_rpc_addr: SocketAddr,
+    config: Config,
```
Ash-L2L, line 628: This should be in `WalletInner`, to preserve the cheap-to-clone property.
Please fix this in a follow-up PR

Ash-L2L (approved):

PR #127, opened 2025-01-15
Fixups for #126

Ash-L2L:
* A broadcast channel is the wrong kind here, since there is only one consumer. Changed to use oneshot channels
* Removed the shutdown function on the wallet, send shutdown notification on drop instead


torkelrogstad (approved):
As always much more idiomatic than me!

PR #134, opened 2025-01-17
update proto, track ctip seq numbers, bump minor version

Ash-L2L:
This is a breaking change to the validator DB.
Upgrading the signet miner server will require:
* stopping the mining script
* saving the wallet db
* syncing a new enforcer (v0.2.0) with a different data dir, etc, in order to get a new validator db
* shutting down the old enforcer (v0.1.x)
* copying over the wallet db

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,7 +494,7 @@ dependencies = [
 
 [[package]]
 name = "bip300301_enforcer"
-version = "0.1.7"
+version = "0.2.0"
 dependencies = [
  "aes-gcm",
  "anyhow",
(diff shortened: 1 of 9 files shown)


torkelrogstad (approved):
LGTM

Re. the change in DB schema: wouldn't it work to wipe the _validator_ data, but leave the _wallet_ data intact? They are two distinct directories, after all

Ash-L2L:
>Re. the change in DB schema: wouldn't it work to wipe the validator data, but leave the wallet data intact? They are two distinct directories, after all

Yes, the wallet data should be kept intact. The old validator data can be wiped after setting up a new validator to sync from the old one.

PR #136, opened 2025-01-21
multi: more trace logs for BMM request, improve wallet locking during sync

torkelrogstad:


Cargo.toml
```diff
@@ -51,9 +51,9 @@ serde_path_to_error = "0.1.16"
 thiserror = "2.0.11"
 tokio = { version = "1.36.0", features = ["rt-multi-thread", "macros"] }
 tokio-stream = "0.1.15"
-tonic = "0.12.3"
-tonic-health = "0.12.3"
-tonic-reflection = "0.12.3"
+tonic = "0.12.2"
```
Ash-L2L, line 54: Why is tonic being downgraded here?
Also, this will not pin the version to `0.12.2`. It will still use `0.12.3`. You need to pin the version if you want to only use `0.12.2`, rather than `0.12.x` where `x >= 2`

src/wallet/mod.rs
```diff
@@ -551,15 +577,22 @@ impl WalletInner {
...
-            wallet_write
-        } else {
-            tracing::trace!("wallet is locked, skipping sync");
-            return Ok(());
+        let wallet_read = match self.read_wallet() {
+            Ok(wallet_read) => wallet_read,
+
+            // "Accepted" errors, that aren't really errors in this case.
+            Err(WalletInitialization::NotUnlocked | WalletInitialization::NotFound) => {
+                tracing::trace!("sync: skipping sync due to wallet error");
```
Ash-L2L, line 585: This should probably be a warning

Ash-L2L (changes requested):

src/wallet/mod.rs
```diff
@@ -601,6 +602,10 @@ impl WalletInner {
             .sync(request, BATCH_SIZE, FETCH_PREV_TXOUTS)
             .into_diagnostic()?;
 
+        // Be a bit smart about the wallet locks, and only acquire the write lock
+        // after the sync itself has completed and we're ready the apply
+        // it to the wallet.
+        let mut wallet_write = self.write_wallet()?;
```
Ash-L2L, line 608: Looks like you might want an upgradeable read lock?
https://docs.rs/lock_api/0.4.12/lock_api/struct.RwLock.html#tymethod.upgradable_read

Ash-L2L (changes requested):

Cargo.toml
torkelrogstad: Aha, that explains why the lock file didn't update. Thanks! This was initial attempt at debugging this by changing Tonic versions. Not the intention to get this into master

src/wallet/mod.rs
torkelrogstad: Why? Will lead to really noisy errors if the enforcer starts with a locked wallet. Every 15s a warning will be printed

Cargo.toml
torkelrogstad: Removed this commit

src/wallet/mod.rs
Ash-L2L: Looks like `lock_api` does not currently support mapping over an `RwLockUpgradableReadGuard`.

Ash-L2L (approved):

PR #137, opened 2025-01-22
use p2p directly to broadcast M5 txs

Ash-L2L:


Cargo.toml
```diff
@@ -64,6 +64,10 @@ zeromq = "0.4.1"
 git = "https://github.com/Ash-L2L/bip300301.git"
 rev = "45f74e37b8295207d29ddbbce10d563ec9f67151"
 
+[dependencies.bitcoin-send-tx-p2p]
+git = "https://github.com/Ash-L2L/bitcoin-send-tx-p2p"
+rev = "4e03ed7c6807771b4709864b0357e2bd8843c83f"
+
```
torkelrogstad, line 70: imo it's a bit unfortunate that we're splitting up functionality over so many different repos. bip300301, cusf-enforcer-mempool, now this as well. what do you think about pulling this into this repo? can do a separate module/file with copyright notices for the original author

src/rpc_client.rs
```diff
@@ -58,3 +61,46 @@ pub fn create_client(
...
+        .send_raw_transaction(encoded_tx, None, Some(MAX_BURN_AMOUNT))
+        .await
+    {
+        Ok(txid) => {
+            tracing::debug!(%txid, "broadcast tx successfully");
+            Ok(Some(txid))
+        }
+        Err(err) => {
+            const OP_DRIVECHAIN_NOT_SUPPORTED_ERR_MSG: &str =
+                "non-mandatory-script-verify-flag (NOPx reserved for soft-fork upgrades)";
```
torkelrogstad, line 96: seems a bit brittle to match exactly on the message. perhaps doing a string contains for `non-mandatory-script-verify-flag` instead?

src/wallet/mod.rs
```diff
@@ -1563,37 +1596,6 @@ impl Wallet {
         }
     }
 
-    // Broadcasts a transaction to the Bitcoin network.
-    pub async fn broadcast_transaction(&self, tx: bdk_wallet::bitcoin::Transaction) -> Result<()> {
-        // Note: there's a `broadcast` method on `bitcoin_blockchain`. We're NOT using that,
-        // because we're broadcasting transactions that "burn" bitcoin (from a BIP-300/1 unaware
```
torkelrogstad, line 1569: I think the continous moving around of code is a bit unfortunate. It makes for noisy diffs, and more importantly it makes diffs harder to review for "real" changes vs. pure moves. If you want to insist on keeping files readable top to bottom I'd appreciate it if you could start doing pure code moves in separate commits, and then do the real changes in separate commits.

Cargo.toml
A-Manning: This is just a fork of https://github.com/andrewtoth/bitcoin-send-tx-p2p
We can use the upstream repo if this PR is merged
https://github.com/andrewtoth/bitcoin-send-tx-p2p/pull/2

src/rpc_client.rs
A-Manning: There are several other errors that use 'non-mandatory-script-verify-flag'. Unfortunately I don't see a better way to do this

src/wallet/mod.rs
A-Manning: Agreed, will do this in future

Cargo.toml
Ash-L2L: This is just a fork of https://github.com/andrewtoth/bitcoin-send-tx-p2p
We can use the upstream repo if this PR is merged
https://github.com/andrewtoth/bitcoin-send-tx-p2p/pull/2

src/wallet/mod.rs
torkelrogstad: Great stuff!

src/rpc_client.rs
Ash-L2L: There are several other errors that use 'non-mandatory-script-verify-flag'. Unfortunately I don't see a better way to do this

src/wallet/mod.rs
Ash-L2L: Agreed, will do this in future

PR #150, opened 2025-02-03
app: add request IDs to req/res/traces

torkelrogstad:
Adds a new ID to all requests/responses. The ID is propagated to logs, under the `request_id` field. The ID can be specified in the request, in the `x-request-id` header. If none is provided, a random UUID is used. The request ID is sent back in the response, in the same header.


torkelrogstad:
Test failure is unrelated, I believe

PR #152, opened 2025-02-04
multi: delay sync task until initial mempool sync is complete

torkelrogstad:


app/main.rs
```diff
@@ -331,6 +331,8 @@ async fn mempool_task<Enforcer, RpcClient, F, Fut>(
         }
     }
 
+    let mempool_sync_span = tracing::span!(tracing::Level::DEBUG, "initial_mempool_sync");
+    let mempool_sync_guard = mempool_sync_span.enter();
```
Ash-L2L, line 335: I'm not sure if this will work as intended across an await. I think the future needs to be instrumented with the span, see https://docs.rs/tracing/0.1.41/tracing/instrument/trait.Instrument.html#method.instrument

lib/wallet/mod.rs
```diff
@@ -580,19 +588,26 @@ impl Task {
...
                         tracing::error!("wallet sync error: {err:#}");
                     }
+                    drop(guard);
                 }
             }
         }
     }
 
-    fn new(wallet: Arc<WalletInner>) -> Self {
+    pub fn new(wallet: Wallet) -> Self {
```
Ash-L2L, line 606: I'm not a fan of this being cyclical - `Task` depends on `Wallet`, which depends on `WalletInner`, which depends on `Task`

Ash-L2L (changes requested):

app/main.rs
torkelrogstad: Fixed, thanks

lib/wallet/mod.rs
torkelrogstad: Good point! Agreed.

torkelrogstad:
Closing in favor of #154 and #155

PR #153, opened 2025-02-04
wallet: up sync batch size

torkelrogstad:


torkelrogstad:
Closing for now, because we're skipping the wallet sync task on the signet server. Can revisit if the problem pops up other places.

PR #154, opened 2025-02-04
Hold write lock on wallet during sync, delay sync until wallet is initialized

Ash-L2L:


lib/wallet/sync.rs
```diff
@@ -67,28 +67,34 @@ impl WalletInner {
...
+        );
         const BATCH_SIZE: usize = 5;
         const FETCH_PREV_TXOUTS: bool = false;
         let update = self
             .electrum_client
             .sync(request, BATCH_SIZE, FETCH_PREV_TXOUTS)?;
+        tracing::trace!("Fetched update from electrum client, applying update");
         // Be a bit smart about the wallet locks, and only acquire the write lock
         // after the sync itself has completed and we're ready the apply
         // it to the wallet.
```
torkelrogstad, line 97: These comments are outdated now

lib/wallet/sync.rs
```diff
@@ -67,28 +67,34 @@ impl WalletInner {
         let start = SystemTime::now();
         tracing::trace!("starting wallet sync");
         // Don't error out here if the wallet is locked, just skip the sync.
-        let wallet_read = match self.read_wallet() {
-            Ok(wallet_read) => wallet_read,
+        let mut wallet_write = match self.write_wallet() {
+            Ok(wallet_write) => wallet_write,
```
torkelrogstad, line 71: I don't really understand why this locking behavior is better than the previous one. Would be good with some explanations in comments as to why we're choosing write/read at different steps.

lib/wallet/mod.rs
```diff
@@ -566,40 +567,73 @@ impl Task {
...
+        use futures::future::{select, Either};
         const SYNC_INTERVAL: Duration = Duration::from_secs(15);
-        tracing::debug!("wallet sync task: starting with interval {SYNC_INTERVAL:?}",);
-
-        let mut interval = interval(SYNC_INTERVAL);
+        tracing::debug!(
+            interval = %humantime::format_duration(SYNC_INTERVAL),
+            "wallet sync task: starting"
+        );
+        shutdown_rx = match select(start_rx, shutdown_rx).await {
```
torkelrogstad, line 581: I had to stare at this for a bit before understanding what's happening. IMO a comment like this would be great. 

`Wait for the start signal to be sent before starting the sync. Take care to also listen for the shutdown signal or channel closures, which might happen while waiting for start`

lib/wallet/cusf_block_producer.rs
```diff
@@ -41,20 +41,30 @@ impl CusfEnforcer for Wallet {
...
-            Ok(())
-        } else {
-            Err(Self::SyncError::WalletTip {
-                expected: tip_hash,
-                wallet_tip,
-            })
-        }
+        tracing::debug!(%tip_hash, "Synced validator, syncing wallet..");
+
+        tokio::task::block_in_place(|| {
```
torkelrogstad, line 46: What does `block_in_place` do here? I don't understand why it's needed

lib/wallet/mod.rs
```diff
@@ -554,6 +553,8 @@ impl WalletInner {
 }
 
 pub struct Task {
+    /// Send a startup signal
+    start_tx: Mutex<Option<oneshot::Sender<()>>>,
```
torkelrogstad, line 557: Perhaps this is a stupid question - but why does this have to be an Option? I tried removing both the mutex and the option, and ran into compiler issues. I understand the Mutex, because of needing to be copyable. When removing the Option I also get complaints about `Copy`. Why is that?

Ash-L2L: This has to be an Option because the sender channel will be consumed when sending the first signal. So after sending the first signal, there is no `oneshot::Sender`.

PR #156, opened 2025-02-06
Use upgradable rwlock for BDK wallet, do not block while initializing wallet

Ash-L2L:


torkelrogstad (approved):
This is way above my pay grade, am afraid I don't really have any input here

PR #159, opened 2025-02-07
Wallet: disable periodic sync by default

Ash-L2L:


torkelrogstad:
Considering the periodic sync is an issue on larger wallets, doesn't it make sense to keep it enabled by default? On a small wallet we're talking sync times on the order of 100ms

PR #168, opened 2025-02-10
wallet: add required_utxos parameter to transaction construction

torkelrogstad:


lib/wallet/mod.rs
```diff
@@ -657,6 +657,17 @@ impl Drop for Task {
     }
 }
 
+/// Optional parameters for sending a wallet transaction
+#[derive(Debug, Default)]
+pub struct SendTransactionParams {
```
Ash-L2L, line 662: This should be renamed to something like `CreateTransactionParams`

Ash-L2L (changes requested):

lib/server.rs
```diff
@@ -1166,8 +1176,27 @@ impl WalletService for crate::wallet::Wallet {
...
             fee_rate,
             op_return_message,
+            required_utxos,
         } = request.into_inner();
 
+        let required_utxos = required_utxos
+            .iter()
+            .map(|utxo| {
+                let txid = utxo.txid.as_ref().ok_or_else(|| {
+                    missing_field::<SendTransactionRequest>("required_utxos.txid")
```
Ash-L2L, line 1186: This will generate a misleading error - the missing field is from `RequiredUtxo`, and is named `txid`

lib/server.rs
```diff
@@ -1166,8 +1176,27 @@ impl WalletService for crate::wallet::Wallet {
...
+                })?;
+
+                let txid = txid
+                    .clone()
+                    .decode_tonic::<SendTransactionRequest, _>("required_utxos.txid")?;
```
Ash-L2L, line 1191: This will generate a misleading error - the decoded field is from `RequiredUtxo`, and is named `txid`

lib/server.rs
```diff
@@ -93,6 +93,14 @@ impl IntoStatus for miette::Report {
             return source.clone().into();
         }
 
+        if let Some(source) = self.downcast_ref::<crate::wallet::error::SendTransaction>() {
```
Ash-L2L, line 96: It is not great to rely on downcasting here. It would be much better to use typed errors instead of using `miette::Report` everywhere. `IntoStatus` could be implemented for typed errors without having to downcast a `miette::Report`

Ash-L2L (changes requested):

lib/server.rs
torkelrogstad: I get your point - but I also have to say I disagree. Might just be me that hasn't fully leaned into the Rust style, but I find it really tiring to have to keep error representations in sync whenever something is changed. The responses here aren't critical, either. They're just used for getting sensible error codes out to the user. 

Would it be okay for you if we kept this as-is for now?

lib/server.rs
torkelrogstad: You're right, this is better!

```
$ just curl -d @request.json http://localhost:50051/cusf.mainchain.v1.WalletService/SendTransaction
{
   "code": "invalid_argument",
   "message": "Missing field in message `cusf.mainchain.v1.SendTransactionRequest.RequiredUtxo`: `txid`"
}
```

lib/server.rs
Ash-L2L: Accepting this for now, it would be quite a lot of work to move away from downcasting

PR #172, opened 2025-02-13
wallet: add support for initiating from seed file

mplatt8:


lib/wallet/mnemonic.rs
```diff
@@ -24,6 +25,22 @@ pub(crate) fn new_mnemonic() -> Result<Mnemonic, miette::Report> {
     Mnemonic::parse(words).into_diagnostic()
 }
 
+/// Read and parse a mnemonic from a file containing exactly 12 space-separated words.
+pub fn new_from_file(path: impl AsRef<Path>) -> Result<Mnemonic, miette::Report> {
```
torkelrogstad, line 29: This should be `pub (crate)`

lib/wallet/mnemonic.rs
```diff
@@ -24,6 +25,22 @@ pub(crate) fn new_mnemonic() -> Result<Mnemonic, miette::Report> {
...
+    let mnemonic_str = std::fs::read_to_string(path)
+        .map_err(|err| miette!("failed to read mnemonic file: {}", err))?;
+
+    let words: Vec<&str> = mnemonic_str.split_whitespace().collect();
+    if words.len() != 12 {
+        return Err(miette!(
+            "mnemonic must be exactly 12 words, got {}",
+            words.len()
+        ));
+    }
```
torkelrogstad, line 39: Is this a hard requirement from our side? Validity of the mnemonic is verified on line 41, and valid mnemonics can be of various lengths

app/main.rs
```diff
@@ -540,11 +540,26 @@ async fn main() -> Result<()> {
...
-            tracing::info!("auto-creating new wallet");
-            let mnemonic = None;
-            let password = None;
-            wallet.create_wallet(mnemonic, password).await?;
+        if !wallet.is_initialized().await {
+            if cli.wallet_opts.auto_create && cli.wallet_opts.mnemonic_path.is_some() {
+                return Err(miette!(
+                    "Cannot use both --wallet-auto-create and --set-seed-with-mnemonic"
+                ));
+            }
```
torkelrogstad, line 548: This can be done with the `conflicts_with` attribute in the config struct instead, with the `conflicts_with` attribute

torkelrogstad (changes requested):

lib/cli.rs
```diff
@@ -253,6 +253,11 @@ pub struct WalletConfig {
     /// the wallet is large and periodic syncs are not feasible.
     #[arg(long = "wallet-skip-periodic-sync", default_value_t = false)]
     pub skip_periodic_sync: bool,
+
+    /// Path to a file containing exactly 12 space-separated BIP39 mnemonic words.
+    /// Cannot be used together with --wallet-auto-create.
+    #[arg(long = "set-seed-with-mnemonic")]
```
torkelrogstad, line 259: I think a better name here would be `wallet-seed-file`. This flag sounds like a bool, and the other wallet flags start with `wallet-`

app/main.rs
```diff
@@ -540,11 +540,26 @@ async fn main() -> Result<()> {
...
+                wallet.create_wallet(Some(mnemonic), None).await?;
+                tracing::info!(
+                    "Successfully created wallet from mnemonic file: {}",
+                    mnemonic_path.display()
+                );
+            } else if cli.wallet_opts.auto_create {
+                tracing::info!("auto-creating new wallet");
+                wallet.create_wallet(None, None).await?;
+            }
         }
```
torkelrogstad, line 563: I think this is better, with only one instance of calling `wallet.create_wallet`

```
  let (mnemonic, auto_create) =
                match (cli.wallet_opts.mnemonic_path, cli.wallet_opts.auto_create) {
                    (Some(mnemonic_path), _) => {
                        tracing::info!("Creating wallet from mnemonic file");
                        let mnemonic =
                            bip300301_enforcer_lib::wallet::mnemonic::new_from_file(mnemonic_path)?;
                        (Some(mnemonic), false)
                    }
                    (_, true) => {
                        tracing::info!("auto-creating new wallet");
                        (None, true)
                    }
                    _ => (None, false),
                };

            wallet.create_wallet(mnemonic, None).await?;
```

PR #173, opened 2025-02-14
wallet: add esplora as chain data source

torkelrogstad:
From my experience Esplora seems a lot more performant. The Electrum server we're using exposes both an Electrum server and an Esplora server, so this works nicely out of the box.


torkelrogstad:
Merging this, and holding off on making it the default until I've been able to figure out the signet server issues. Syncing my local wallet is just fine with Esplora, so there's probably some shenanigans related to the size of the signet server.

PR #184, opened 2025-03-07
Add support for rolling logs

Ash-L2L:


lib/cli.rs
```diff
@@ -168,13 +183,29 @@ impl<'writer> tracing_subscriber::fmt::FormatFields<'writer> for LogFormatter {
...
+    fn from(rotation: LogRotation) -> Self {
+        match rotation {
+            LogRotation::Daily => Self::DAILY,
+            LogRotation::Hourly => Self::HOURLY,
+            LogRotation::Minutely => Self::MINUTELY,
+            LogRotation::Never => Self::NEVER,
+        }
+    }
+}
+
```
torkelrogstad, line 206: is it hard to implement log rotation based on log file size?

app/Cargo.toml
```diff
@@ -25,10 +24,15 @@ tonic-reflection = "0.12.3"
...
 tracing = { workspace = true }
-tracing-appender = "0.2.3"
+tracing-appender = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 uuid = "1.12.1"
 
+[dependencies.bip300301_enforcer_lib]
+path = "../lib"
+default-features = false
+features = ["tracing-appender"]
```
torkelrogstad, line 34: Do we really need this feature flag? If it's only for reducing the single `tracing-appender` lib I don't think it's worth it. Feature flags make everything a bit harder to understand when stumbling across the code, IMO

Ash-L2L: Removed the feature flag

PR #185, opened 2025-03-07
Ensure that events are broadcast only after committing db txs

Ash-L2L:
This seemingly minor change fixes an observed bug on signet:
Sidechain's requests for headers are hitting the Enforcer only 2550μs after broadcasting an event.
However, committing the db txs after broadcasting an event is taking slightly longer (unknown, but less than 6970μs)
This causes the Enforcer to respond that a block header is missing.

Fixed by broadcasting events only after committing state changes

### DIFF
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -590,9 +590,8 @@ pub fn validate_tx(
 pub(in crate::validator) fn connect_block(
     rwtxn: &mut RwTxn,
     dbs: &Dbs,
-    event_tx: &Sender<Event>,
     block: &Block,
-) -> Result<(), error::ConnectBlock> {
+) -> Result<Event, error::ConnectBlock> {
     let parent = block.header.prev_blockhash;
 
     // Check that current chain tip is block parent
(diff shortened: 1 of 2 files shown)


torkelrogstad (approved):
Good catch!

lib/validator/task/mod.rs
```diff
@@ -855,9 +853,10 @@ where
         let mut rwtxn = dbs.write_txn()?;
         let height = dbs.block_hashes.height().get(&rwtxn, &missing_block)?;
         // FIXME: handle disconnects
-        let () = connect_block(&mut rwtxn, dbs, event_tx, &block)?;
+        let event = connect_block(&mut rwtxn, dbs, &block)?;
         tracing::debug!("connected block at height {height}: {missing_block}");
         let () = rwtxn.commit()?;
+        let _send_err: Result<Option<_>, TrySendError<_>> = event_tx.try_broadcast(event);
```
torkelrogstad, line 859: Could consider leaving a comment here about it being important that we commit before sending the event, referencing this PR

PR #186, opened 2025-03-08
rpc: add support for CreateMultisigAddress and CreateMultisigTransaction

mplatt8:


lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+    /// Returns a tuple containing:
+    /// * The multisig address
+    /// * The redeem script in hex
+    /// * The output descriptor
+    pub fn create_multisig_address(
+        &self,
+        required_signatures: u32,
+        public_keys: &[String],
+        address_type: Option<&str>,
+    ) -> Result<(bitcoin::Address, String, String)> {
```
torkelrogstad, line 1952: I'm not sure the correct place for this is in the enforcer /wallet/. It doesn't actually use any parts of the wallet, from what I can tell. Would the validator be a better place?

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+            return Err(miette!(
+                "required_signatures ({}) cannot be greater than the number of public keys ({})",
+                required_signatures,
+                public_keys.len()
+            ));
+        }
+
+        if public_keys.is_empty() {
+            return Err(miette!("public_keys cannot be empty"));
+        }
```
torkelrogstad, line 1968: This should be validated in `server.rs` IMO, in a way that proper tonic statuses with the `InvalidArgument` status code is returned.

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+    ///
+    /// # Returns
+    /// Returns a tuple containing:
+    /// * The multisig address
+    /// * The redeem script in hex
+    /// * The output descriptor
+    pub fn create_multisig_address(
+        &self,
+        required_signatures: u32,
+        public_keys: &[String],
```
torkelrogstad, line 1950: Better to take in a list of `bitcoin::key::PublicKey` than strings

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+        address_type: Option<&str>,
```
torkelrogstad, line 1951: Should use an enum here, not a `str`

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+            "p2sh" => {
+                let redeem_script = script.clone();
+                // p2sh returns a Result<Address, P2shError>
+                let address = match bitcoin::Address::p2sh(&redeem_script, network) {
+                    Ok(addr) => addr,
+                    Err(_) => return Err(miette!("failed to create p2sh address")),
+                };
+
+                let descriptor = format!(
+                    "sh(multi({},{}))#multisig",
```
torkelrogstad, line 2016: The descriptor checksums here are invalid

lib/wallet/mod.rs
torkelrogstad: We should also not be returning strings. For the descriptor, we should be returning a properly typed descriptor from the `miniscript` crate: https://docs.rs/miniscript/12.3.0/miniscript/index.html#output-descriptors. It is re-exported by `bdk_wallet`. For the redeem script, what about a `ScriptBuf`?

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+    /// * `utxos` - List of UTXOs to include in the transaction
+    /// * `fee_policy` - Optional fee policy to use
+    /// * `op_return_message` - Optional OP_RETURN message to include
+    ///
+    /// # Returns
+    /// A tuple containing the PSBT and transaction ID
+    pub fn create_multisig_transaction(
+        &self,
+        multisig_address: &str,
+        redeem_script_hex: &str,
```
torkelrogstad, line 2098: Seme here as above: we should avoid string types

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+        let network = bitcoin::Network::from_str(validator_network.to_string().as_str())
+            .map_err(|e| miette!("Failed to convert network: {}", e))?;
+
+        let multisig_addr = bitcoin::Address::from_str(multisig_address)
+            .map_err(|_| miette!("invalid multisig address: {}", multisig_address))?;
+
+        // Verify the address is on the correct network
+        let address = multisig_addr
+            .require_network(network)
+            .map_err(|_| miette!("address network mismatch: expected {}", network))?;
```
torkelrogstad, line 2115: Should be validated in the outer layer, with proper tonic status codes

lib/wallet/mod.rs
```diff
@@ -1931,4 +1931,280 @@ impl Wallet {
...
+
+        let redeem_script = bitcoin::ScriptBuf::from_bytes(redeem_script_bytes);
+
+        // Create a new transaction
+        let mut tx = bitcoin::Transaction {
+            version: bitcoin::transaction::Version::TWO,
+            lock_time: bitcoin::absolute::LockTime::ZERO,
+            input: Vec::new(),
+            output: Vec::new(),
+        };
```
torkelrogstad, line 2129: I'm very skeptical towards building the transaction this way, I have to say. What about the BDK transaction builder?

torkelrogstad: Without using the transaction builder, there's no point in this function being in the wallet, either

torkelrogstad (commented):
The `cusf_sidechain_proto` git commit in the sub repo can be updated to the commit of your Proto PR, such that this is possible to compile

PR #190, opened 2025-03-31
Update to latest proto (header sync progress stream)

seriouscatsoserious:
- Add HeaderSyncProgress and SubscribeHeaderSync proto definitions
- Implement header sync streaming in validator
- Add subscribe_header_sync RPC endpoint with progress tracking
- Update submodule with proto changes

### DIFF
--- a/lib/server.rs
+++ b/lib/server.rs
@@ -175,16 +175,6 @@ impl ValidatorService for Validator {
                 match rx.changed().await {
                     Ok(()) => {
                         let progress = rx.borrow().clone();
-                        if let (Some(current), Some(target)) =
-                            (progress.current_height, progress.target_height)
-                        {
-                            tracing::debug!(
-                                "Sending header sync progress: {}/{} ({:.1}%)",
-                                current,
-                                target,
-                                (current as f64 / target as f64) * 100.0
-                            );
-                        }
                         Ok(Some((progress.into(), rx)))
                     }
                     Err(_) => Ok(None),
--- a/lib/server.rs
+++ b/lib/server.rs
@@ -53,8 +53,9 @@ use crate::{
             ListSidechainDepositTransactionsRequest, ListSidechainDepositTransactionsResponse,
             ListTransactionsRequest, ListTransactionsResponse, ListUnspentOutputsRequest,
             ListUnspentOutputsResponse, Network, SendTransactionRequest, SendTransactionResponse,
-            SubscribeEventsRequest, SubscribeEventsResponse, UnlockWalletRequest,
-            UnlockWalletResponse, WalletTransaction,
+            SubscribeEventsRequest, SubscribeEventsResponse, SubscribeHeaderSyncRequest,
+            SubscribeHeaderSyncResponse, UnlockWalletRequest, UnlockWalletResponse,
+            WalletTransaction,
         },
         IntoStatus,
     },
@@ -160,6 +161,39 @@ impl IntoStatus for miette::Report {
 
 #[tonic::async_trait]
 impl ValidatorService for Validator {
+    type SubscribeHeaderSyncStream =
+        BoxStream<'static, Result<SubscribeHeaderSyncResponse, tonic::Status>>;
+
+    async fn subscribe_header_sync(
+        &self,
+        _request: tonic::Request<SubscribeHeaderSyncRequest>,
+    ) -> Result<tonic::Response<Self::SubscribeHeaderSyncStream>, tonic::Status> {
+        let rx = self.subscribe_header_sync();
+        let initial = rx.borrow().clone();
+        let stream = futures::stream::once(async { Ok(initial.into()) })
+            .chain(futures::stream::try_unfold(rx, |mut rx| async move {
+                match rx.changed().await {
+                    Ok(()) => {
+                        let progress = rx.borrow().clone();
+                        if let (Some(current), Some(target)) =
+                            (progress.current_height, progress.target_height)
+                        {
+                            tracing::debug!(
+                                "Sending header sync progress: {}/{} ({:.1}%)",
+                                current,
+                                target,
+                                (current as f64 / target as f64) * 100.0
+                            );
+                        }
+                        Ok(Some((progress.into(), rx)))
+                    }
+                    Err(_) => Ok(None),
+                }
+            }))
+            .boxed();
+        Ok(tonic::Response::new(stream))
+    }
+
     async fn get_block_header_info(
         &self,
         request: tonic::Request<GetBlockHeaderInfoRequest>,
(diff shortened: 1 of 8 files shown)


lib/validator/mod.rs
```diff
@@ -178,6 +179,14 @@ pub enum EventsStreamError {
     Overflow,
 }
 
+#[derive(Debug, Diagnostic, Error)]
+pub enum HeaderSyncStreamError {
+    #[error("Header sync stream closed due to overflow")]
+    Overflow,
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error + Send + Sync>),
```
Ash-L2L, line 187: Why do you need this case?

lib/validator/mod.rs
```diff
@@ -190,7 +199,11 @@ pub enum GetSidechainsError {
 pub struct Validator {
     dbs: Dbs,
     events_rx: InactiveReceiver<Event>,
-    events_tx: Sender<Event>,
+    events_tx: BroadcastSender<Event>,
+    header_sync: Option<(
```
Ash-L2L, line 203: Please give this a better name like `header_sync_progress_channel` or equivalent

lib/validator/mod.rs
```diff
@@ -230,6 +253,20 @@ impl Validator {
         .fuse()
     }
 
+    pub fn subscribe_header_sync(&self) -> watch::Receiver<SubscribeHeaderSyncResponse> {
+        match &self.header_sync {
+            Some((_, rx)) => rx.clone(),
+            None => {
+                // Return an empty receiver if no sync in progress
```
Ash-L2L, line 260: Why? You should return an `Option` to indicate that there is no sync in progress.

This entire method could be

```rust
pub fn subscribe_header_sync(&self) -> Option<&watch::Receiver<SubscribeHeaderSyncResponse>> {
    self.header_sync.map(|(_, rx)| rx)
}
```

lib/validator/task/error.rs
```diff
@@ -226,8 +226,14 @@ pub(in crate::validator) enum ConnectBlock {
 #[derive(Debug, Error)]
 pub(in crate::validator) enum DisconnectBlock {}
 
+#[derive(Debug, Error)]
+#[error("Header sync already in progress")]
+pub struct HeaderSyncInProgressError;
```
Ash-L2L, line 231: Please don't use the `Error` suffix in a module called `error`, it is already clear that this is an error enum

Ash-L2L, line 231: If this error type is only used in one place, then it should just be a case of `Sync`, no need to create a new error type

lib/validator/task/mod.rs
```diff
@@ -767,15 +767,39 @@ pub(in crate::validator) fn disconnect_block(
...
+        .map_err(|err| error::Sync::JsonRpc {
+            method: "getblockheader".to_owned(),
+            source: err,
+        })
+        .await?
+        .height as u64;
+
+    // Send initial progress
+    let progress = SubscribeHeaderSyncResponse {
+        current_height: Some(0),
```
Ash-L2L, line 795: These defaults look wrong

lib/validator/task/mod.rs
```diff
@@ -866,13 +913,24 @@ where
...
     MainClient: bip300301::client::MainClient + Sync,
 {
-    let () = sync_headers(dbs, main_client, main_tip).await?;
-    let () = sync_blocks(dbs, event_tx, main_client, main_tip).await?;
-    Ok(())
+    let Some((tx, _)) = header_sync else {
+        return Err(error::Sync::HeaderSyncInProgress(HeaderSyncInProgressError));
+    };
+
+    // Do sync work
```
Ash-L2L, line 930: This block does the same as it did before, but now requires 5 lines instead of 3. Why change it?

lib/server.rs
```diff
@@ -160,6 +161,39 @@ impl IntoStatus for miette::Report {
...
+                    Ok(()) => {
+                        let progress = rx.borrow().clone();
+                        if let (Some(current), Some(target)) =
+                            (progress.current_height, progress.target_height)
+                        {
+                            tracing::debug!(
+                                "Sending header sync progress: {}/{} ({:.1}%)",
+                                current,
+                                target,
+                                (current as f64 / target as f64) * 100.0
```
Ash-L2L, line 185: Please remove the `f64` part

lib/server.rs
```diff
@@ -160,6 +161,39 @@ impl IntoStatus for miette::Report {
...
+    type SubscribeHeaderSyncStream =
+        BoxStream<'static, Result<SubscribeHeaderSyncResponse, tonic::Status>>;
+
+    async fn subscribe_header_sync(
+        &self,
+        _request: tonic::Request<SubscribeHeaderSyncRequest>,
+    ) -> Result<tonic::Response<Self::SubscribeHeaderSyncStream>, tonic::Status> {
+        let rx = self.subscribe_header_sync();
+        let initial = rx.borrow().clone();
+        let stream = futures::stream::once(async { Ok(initial.into()) })
```
Ash-L2L, line 173: The stream here is quite straightforward, you should just construct a [`WatchStream`](https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.WatchStream.html) and map over it

Ash-L2L (changes requested):

lib/validator/task/mod.rs
```diff
@@ -767,15 +767,39 @@ pub(in crate::validator) fn disconnect_block(
     Ok(())
 }
 
+use crate::validator::task::error::HeaderSyncInProgressError;
```
Ash-L2L, line 770: Please keep imports grouped at the top of the file. Imports in Rust are module-scoped, so this applies to the entire module, not just the code below.

lib/validator/task/mod.rs
```diff
@@ -866,13 +913,24 @@ where
...
     main_client: &MainClient,
     main_tip: BlockHash,
 ) -> Result<(), error::Sync>
 where
     MainClient: bip300301::client::MainClient + Sync,
 {
-    let () = sync_headers(dbs, main_client, main_tip).await?;
-    let () = sync_blocks(dbs, event_tx, main_client, main_tip).await?;
-    Ok(())
+    let Some((tx, _)) = header_sync else {
```
Ash-L2L, line 926: This is wrong. I'm not sure how the `header_sync` arg is supposed to represent sync progress here. If there is already a sync in progress, then why call this function at all?

This function should require a sync progress channel, instead of an optional channel.
In `validator::cusf_enforcer::sync_to_tip`, you can handle initializing the sync progress channel.

Also, it would be good to use a better name than `tx` for the sender, particularly because there is also another channel sender argument (`event_tx`).

Ash-L2L (changes requested):

Ash-L2L (approved):

PR #197, opened 2025-04-17
validator: fetch block headers via REST

### LINKED ISSUE #195: performance: consider fetching headers using REST API
Block headers can be fetched in batches when done over the REST API, whereas JSON-RPC (what we're currently using) only allows fetching a single one. Fetching headers using batched JSON-RPC is not possible, as we only know hash N-1 after having successfully fetched hash N.

Docs for this is here: https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#blockheaders

torkelrogstad:
Notes: this relies on https://github.com/Ash-L2L/bip300301/pull/5 getting merged. 

Closes #195 

This is tested with a few different reorg scenarios on regtest. I've done tests with actual reorgs (two nodes running, where one overtakes the other), as well as by doing RPC calls to `invalidateblock`. Note that calling `invalidateblock` leads to a TODO panic within `cusf-enforcer-mempool`, but that is unrelated to this PR. 

The implementation here is rather chaotic and messy. I've tried my best to comment things liberally, but I'm very much open to suggestions on how to simplify. 

Stats for syncing headers on signet, on my Macbook Pro with Bitcoin Core running locally: 

* ~This PR: 5.05525225s~
* This PR after reorg logic integrated: 6.464624708s
* master: 47.817417791s

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ rev = "543681a65f8f9b44b298482a7f39dd688829a2de"
 
 [workspace.dependencies.cusf-enforcer-mempool]
 git = "https://github.com/LayerTwo-Labs/cusf-enforcer-mempool.git"
-rev = "4730a57b9f9d84779174f061ad704eea317409b2"
+rev = "e041fd65a92f64baddb98689f3b9da3b1c28c61b"
 
 [workspace.dependencies.ouroboros]
 git = "https://github.com/erikjohnston/ouroboros.git"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,11 +2,7 @@
 
 [workspace]
 resolver = "2"
-members = [
-    "app",
-    "lib",
-    "integration_tests",
-]
+members = ["app", "lib", "integration_tests"]
 
 [workspace.package]
 authors = [
@@ -27,6 +23,7 @@ futures = "0.3.30"
 hex = "0.4.3"
 jsonrpsee = { version = "0.24.7" }
 miette = { version = "7.1.0", default-features = false }
+reqwest = { version = "0.12.15", default-features = false, features = ["json"] }
 thiserror = "2.0.11"
 tokio = { version = "1.36.0", default-features = false }
 tokio-stream = "0.1.15"
@@ -37,12 +34,11 @@ tracing-subscriber = "0.3.18"
 
 [workspace.dependencies.bip300301]
 git = "https://github.com/Ash-L2L/bip300301.git"
-rev = "45f74e37b8295207d29ddbbce10d563ec9f67151"
+rev = "543681a65f8f9b44b298482a7f39dd688829a2de"
 
 [workspace.dependencies.cusf-enforcer-mempool]
-# https://github.com/LayerTwo-Labs/cusf-enforcer-mempool/pull/5
-git = "https://github.com/torkelrogstad/cusf-enforcer-mempool.git"
-rev = "26bb6bdeb6cb9a7d5222889f49bb1768ccc2e011"
+git = "https://github.com/LayerTwo-Labs/cusf-enforcer-mempool.git"
+rev = "4730a57b9f9d84779174f061ad704eea317409b2"
 
 [workspace.dependencies.ouroboros]
 git = "https://github.com/erikjohnston/ouroboros.git"
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, cmp::Ordering, collections::HashMap};
+use std::{borrow::Cow, cmp::Ordering, collections::HashMap, time::Instant};
 
 use async_broadcast::{Sender, TrySendError};
 use bip300301::client::{GetBlockClient, U8Witness};
@@ -27,6 +27,8 @@ use crate::{
     },
     validator::dbs::{db_error, Dbs, RwTxn, UnitKey},
 };
+
+use super::main_rest_client::MainRestClient;
 
 pub mod error;
 
@@ -151,7 +153,9 @@ fn handle_failed_sidechain_proposals(
         .map_err(db_error::Iter::from)?
         .map_err(|err| error::HandleFailedSidechainProposals::DbIter(err.into()))
         .filter_map(|(proposal_id, sidechain)| {
-            let sidechain_proposal_age = height - sidechain.status.proposal_height;
+            // doing `height - sidechain.status.proposal_height` can panic if the enforcer has data
+            // from a previous sync that is not in the active chain.
+            let sidechain_proposal_age = height.saturating_sub(sidechain.status.proposal_height);
             let sidechain_slot_is_used = dbs
                 .active_sidechains
                 .sidechain()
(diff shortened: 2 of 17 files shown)


lib/validator/task/mod.rs
```diff
@@ -768,98 +770,167 @@ pub(in crate::validator) fn disconnect_block(
     todo!();
 }
 
-async fn sync_headers<MainClient>(
+#[allow(clippy::result_large_err)]
```
Ash-L2L, line 773: If this is needed, it is best to figure out which error variant is large and box it, instead of ignoring the lint

lib/validator/task/mod.rs
```diff
@@ -768,98 +770,167 @@ pub(in crate::validator) fn disconnect_block(
...
-        } else {
-            tracing::debug!("Syncing header `{latest_missing_header}` -> `{main_tip}`");
+    let start = Instant::now();
+
+    let (current_enforcer_tip, current_enforcer_tip_height) = match fetch_chain_tip(dbs)? {
+        Some(tip) => tip,
+        None => {
+            let genesis_hash =
+                main_rpc_client
+                    .getblockhash(0)
```
Ash-L2L, line 803: This is not correct -
* If the enforcer does not have a chain tip, that does not imply that the enforcer is synced to the genesis block in the active chain.
* Fetching blocks by height is almost always a bad idea - what if the active chain changes?

lib/validator/task/mod.rs
```diff
@@ -768,98 +770,167 @@ pub(in crate::validator) fn disconnect_block(
...
+    // The requested block header is the /first/ header in the batch. This means we
+    // need to loop from our current tip, until we find the requested main tip.
+    loop {
+        tracing::debug!(
+            "Fetching batch of headers starting from #{current_height} `{current_block_hash}`"
+        );
+
+        // Fetch a batch of headers
+        let headers = main_rest_client
+            .get_block_headers(&current_block_hash, HEADER_FETCH_BATCH_SIZE)
```
Ash-L2L, line 833: Can you be certain that the current enforcer tip is in the active chain?

Ash-L2L (changes requested):

lib/validator/task/mod.rs
torkelrogstad: I'm not sure I follow here. If there's no chain tip - what does that imply?

My idea here is that if there's no chain tip, we need to start from the very beginning, i.e. genesis. Note that this is the only place we're fetching a block by height. The genesis block will never change, so I don't see how that is problematic

lib/validator/task/mod.rs
```diff
@@ -768,101 +772,387 @@ pub(in crate::validator) fn disconnect_block(
...
+        let ancestor_headers = dbs
+            .block_hashes
             .ancestor_headers(&rotxn, main_tip)
-            .map(|(block_hash, _header)| Ok(block_hash))
-            .take_while(|block_hash| Ok(!dbs.block_hashes.contains_block(&rotxn, block_hash)?))
-            .collect()
-            .map_err(error::Sync::from)
+            .collect::<Vec<_>>()?;
+
+        tracing::info!(
```
A-Manning, line 1089: Can't you just get the block height? IIUC the headers DB always contains all ancestors of any block hash that it contains

lib/validator/task/mod.rs
```diff
@@ -768,101 +772,387 @@ pub(in crate::validator) fn disconnect_block(
...
+            let new_chain_tip = block.header.prev_blockhash;
+            let current_tip = dbs.current_chain_tip.try_get(&rwtxn, &UnitKey)?;
+
+            if current_tip != Some(new_chain_tip) {
+                tracing::info!(
+                    "Syncing first block, updating enforcer tip from {} to {}",
+                    current_tip.unwrap_or(BlockHash::all_zeros()),
+                    new_chain_tip,
+                );
+                dbs.current_chain_tip
```
A-Manning, line 1146: This is updated by `connect_block`, should not be done here

Cargo.toml
```diff
@@ -27,6 +23,9 @@ futures = "0.3.30"
 hex = "0.4.3"
 jsonrpsee = { version = "0.24.7" }
 miette = { version = "7.1.0", default-features = false }
+# Default features include TLS stuff, which on Linux requires OpenSSL. 
+# We only support Bitcoin Core over HTTP, so need for that.
```
A-Manning, line 27: You don't need a comment for this. Disabling default features is always fine if it works

A-Manning (changes requested):

lib/validator/task/mod.rs
torkelrogstad: I added this because it was a point needed in order to sync correctly after a reorg had taken place. When I test now I'm able to do without it... Have to admit this is a bit of a clusterfuck... Removing!

lib/validator/task/mod.rs
torkelrogstad: I'm still not sure how to handle this completely correctly. If we want to fetch block hashes starting at genesis, we need the genesis block hash in order to call the REST API. What do you suggest here, if you think I'm doing something incorrectly?

lib/validator/task/mod.rs
```diff
@@ -804,6 +804,23 @@ fn fetch_chain_tip_enforcer_db(dbs: &Dbs) -> Result<Option<(BlockHash, u32)>, Bo
     Ok(Some((tip, tip_height)))
 }
 
+// This is its own function due to issues with DB txns being sent across futures.
```
Ash-L2L, line 810: You can drop the read txn in order to avoid holding it across an await. eg.
```rust
async fn do_async_stuff() {
   do_something().await;
   // the DB txn will be dropped at the end of this scope
   {
       let rotxn = ...;
       // use the DB txn here
   }
   do_something_else().await
}
```

Ash-L2L (changes requested):

lib/validator/task/mod.rs
```diff
@@ -768,101 +772,387 @@ pub(in crate::validator) fn disconnect_block(
...
     if missing_blocks.is_empty() {
+        tracing::info!("No missing blocks, skipping sync");
         return Ok(());
     }
+    tracing::info!("Syncing {} missing blocks", missing_blocks.len(),);
+    let mut total_blocks_fetched = 0;
     for missing_block in missing_blocks.into_iter().rev() {
-        tracing::debug!("Syncing block `{missing_block}` -> `{main_tip}`");
-        let block = main_client
+        // If we're processing the very first missing block, that means we need to
```
Ash-L2L: This doesn't look right - You should not need to update the stored chain tip here at all. You should
* compute the common ancestor of the enforcer tip and `main_tip`
* disconnect blocks until you reach the common ancestor
* connect blocks that the enforcer already has that are ancestors of `main_tip`
* fetch & connect the missing blocks

Ash-L2L (changes requested):

lib/validator/task/mod.rs
torkelrogstad: bccdc8d65ea8d62f0aa775d5b3f9b231b02d4a14

lib/validator/task/mod.rs
torkelrogstad: 20a87efcfcd6da61886689a08d600843de030e40

lib/validator/task/mod.rs
torkelrogstad: e3440073d1d8176e2e74b62ceab5de5cf71a52f7

lib/validator/task/mod.rs
torkelrogstad: cff13e3fd24b806fd9bf9c62ccf0046374b1b18d

Ash-L2L (approved):

PR #204, opened 2025-05-02
dev: check import sorts with `cargo fmt`

torkelrogstad:
We have to use the nightly Rust version for this, as the relevant options in `rustfmt.toml` are not stable. They have been very much stable (!) in this situation for years, but there are some outstanding issues that prevent it from making its way into stable releases.


Ash-L2L (changes requested):
Should probably remove `rustfmt` from the stable toolchain

.github/workflows/check_lint_build_release.yaml
```diff
@@ -19,6 +19,13 @@ jobs:
...
+      # Needed for `cargo fmt` checks
+      - name: Install latest nightly toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+          components: rustfmt
+
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
```
Ash-L2L, line 31: Should probably remove rustfmt from the stable toolchain

Ash-L2L (changes requested):

lib/proto.rs
```diff
@@ -599,8 +599,9 @@ pub mod mainchain {
 
     impl From<&crate::types::WithdrawalBundleEventKind> for withdrawal_bundle_event::event::Event {
         fn from(event_kind: &crate::types::WithdrawalBundleEventKind) -> Self {
-            use crate::types::WithdrawalBundleEventKind;
```
Ash-L2L, line 602: Strange that these get separated into two groups - I guess rustfmt can't know thatwithdrawal_bundle_event is not an external crate.

lib/validator/cusf_enforcer.rs
```diff
@@ -26,8 +27,6 @@ use crate::{
     },
 };
 
-pub use task::error::ValidateTransaction as ValidateTransactionError;
```
Ash-L2L, line 29: This should be merged into the `use crate::` import, not the external crates group

Ash-L2L (changes requested):

Ash-L2L:
The behavior around local module imports is quite surprising. There is a nightly-only lint that ensures that local module imports are not unintentionally mistaken for crate imports when regrouping
https://github.com/rust-lang/rust/issues/138299

Maybe we should use nightly for clippy lints too?

Ash-L2L (approved):

PR #210, opened 2025-05-05
multi: implement Stop endpoint

torkelrogstad:


lib/validator/mod.rs
```diff
@@ -257,6 +263,25 @@ impl Validator {
         self.network
     }
 
+    pub fn shutdown(&self) {
+        tracing::info!("shutting down validator");
+        let tx = self.shutdown_tx.clone();
```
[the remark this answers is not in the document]
torkelrogstad: You're correct!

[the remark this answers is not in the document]
torkelrogstad: With `&self.shutdown_tx` I get `borrowed data escapes outside of method`

Ash-L2L:
This is an odd change to the `Validator` API IMO - IIUC, the shutdown signal is not handled by the `Validator` task at all, and actually shuts down the entire application.

IMO, it would be best if the validator task (https://github.com/LayerTwo-Labs/cusf-enforcer-mempool/blob/master/lib/cusf_enforcer.rs#L254) accepted a shutdown signal.
Instead of subscribing to the shutdown signal emitted from the validator, the main task should just exit if the validator task completes [here](https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/1243b0b254deb3c3748113163f734fc538d1813c/app/main.rs#L695C10-L695C15), perhaps after waiting a little to ensure that any relevant errors are received before exiting.

torkelrogstad:
> the shutdown signal is not handled by the Validator task at all, and actually shuts down the entire application

That's correct. If we shut down the validator there's no need for anything else to run, so IMO that _is_ shutting down the entire application

I did it this way to mimic how we handled ctrl+c/SIGINT. Trying out your suggestion now!

app/main.rs
```diff
@@ -852,29 +862,67 @@ async fn main() -> Result<()> {
...
+        // The JoinHandle itself resolves to Result<InnerResult, JoinError>
+        // We need to flatten this for the shared future.
+        match res {
+            Ok(Ok(())) => Ok(()),
+            Ok(Err(err)) => Err(err),
+            Err(join_error) => Err(miette!(
+                "main task panicked or was cancelled: {join_error:#}"
+            )),
+        }
+    });
```
torkelrogstad, line 879: This is currently non-shared future. I want to make this cloneable, so I can pass it into `handles` on line 881, as well as selecting on it on line 918. Unable to get the compiler to let me do so...

app/main.rs
```diff
@@ -581,25 +607,23 @@ async fn task(
...
                     }
                 };
                 let sample_block_template =
                     match get_block_template(&mainchain_client, network).await {
                         Ok(block_template) => block_template,
                         Err(err) => {
-                            tracing::error!("failed to get sample block template: {err:#}");
-                            return;
+                            let err = miette::Report::from_err(err);
+                            return Err(err.wrap_err("failed to get sample block template"));
```
Ash-L2L, line 634: Nice, I didn't know about `miette::Report::wrap_err`

lib/validator/task/error.rs
```diff
@@ -231,6 +231,8 @@ pub(in crate::validator) enum DisconnectBlock {}
 
 #[fatality(splitable)]
 pub(in crate::validator) enum Sync {
+    #[error("Shutdown signal received")]
```
Ash-L2L, line 234: Should this be a fatal error?

lib/validator/task/mod.rs
```diff
@@ -980,12 +980,13 @@ where
 }
 
 #[tracing::instrument(skip_all)]
-async fn sync_headers<MainRpcClient>(
+async fn sync_headers<MainRpcClient, Signal: Future<Output = ()> + Send>(
```
Ash-L2L, line 983: Can this constraint on `Signal` go with the other constraints in the `where` clause?

lib/validator/task/mod.rs
```diff
@@ -1144,11 +1156,12 @@ where
 
 // MUST be called after `sync_headers`.
 #[tracing::instrument(skip_all)]
-async fn sync_blocks<MainRpcClient>(
+async fn sync_blocks<MainRpcClient, Signal: Future<Output = ()> + Send>(
```
Ash-L2L, line 1159: Can this constraint on `Signal` go with the other constraints in the `where` clause?

lib/validator/task/mod.rs
```diff
@@ -1211,25 +1235,37 @@ where
     Ok(())
 }
 
-pub(in crate::validator) async fn sync_to_tip<MainClient>(
+pub(in crate::validator) async fn sync_to_tip<MainClient, Signal: Future<Output = ()> + Send>(
```
Ash-L2L, line 1238: Can this constraint on `Signal` go with the other constraints in the `where` clause?

lib/validator/task/error.rs
torkelrogstad: Yes!

Ash-L2L (approved):

PR #213, opened 2025-05-08
Use specific error types throughout wallet

Ash-L2L:


lib/proto.rs
```diff
@@ -4,8 +4,102 @@ use thiserror::Error;
...
+    {
+        Self {
+            code: source.code,
+            fmt_message: Box::new(move |f| std::fmt::Display::fmt(&err_msg, f)),
+            source: Some(either::Left(Box::new(source))),
+        }
+    }
+
+    /// Full status message, including source errors in alternate mode
+    fn status_message(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
torkelrogstad, line 66: Really nice 👌

lib/server.rs
```diff
@@ -84,82 +83,6 @@ where
     crate::proto::Error::missing_field::<Message>(field_name).into()
 }
 
-// The idea here is to centralize conversion of lower layer errors into something meaningful
-// out from the API.
-//
-// Lower layer errors that return `miette::Report` can be easily turned into a meaningful
-// API response by just doing `into_status()`. We also get the additional benefit of a singular
-// place to add logs for unexpected errors.
-impl IntoStatus for miette::Report {
```
torkelrogstad, line 93: Agreed that your take on this is much better than my old stuff. But do we keep the behavior I implemented here?

lib/util.rs
```diff
@@ -0,0 +1,26 @@
+//! Utility types and functions
```
torkelrogstad, line 1: I have to say I really dislike general "util" files. They're always indicative of either poor naming or poor structure. What do you think about calling the file `display.rs` and the struct just `Fn`? That way we'd end up with `crate::display::Fn::new`. Or just move DisplayFn to the single place where it is used.

lib/wallet/error.rs
```diff
@@ -25,6 +27,16 @@ pub struct Electrum {
     message: String,
 }
 
+impl ToStatus for Electrum {
```
torkelrogstad, line 30: I think we can actually delete `Electrum` in its entirety. It's a holdover from past usage that no longer exists

lib/wallet/error.rs
```diff
@@ -144,6 +190,249 @@ pub enum WalletSync {
...
+impl ToStatus for ParseMnemonic {
+    fn builder(&self) -> StatusBuilder {
+        StatusBuilder::new(self)
+    }
+}
+
+#[derive(Debug, Diagnostic, Error)]
+pub enum ReadDbMnemonic {
+    #[error(
+        "invalid mnemonic DB state (plaintext_mnemonic=`{}`, iv=`{}`, ciphertext=`{}`, key_salt=`{}`)",
```
torkelrogstad, line 207: it seems bad to include the plaintext mnemonic here?

lib/wallet/error.rs
```diff
@@ -144,6 +190,249 @@ pub enum WalletSync {
...
+        plaintext_mnemonic: String,
+        iv: String,
+        ciphertext: String,
+        key_salt: String,
+    },
+    #[error(transparent)]
+    ParseMnemonic(#[from] ParseMnemonic),
+    #[error("failed to read mnemonic from DB")]
+    ReadMnemonic(#[source] rusqlite::Error),
+    #[error("rustqlite error")]
```
torkelrogstad, line 223: `rusqlite`, not `rustqlite`

lib/wallet/error.rs
```diff
@@ -144,6 +190,249 @@ pub enum WalletSync {
...
+    MnemonicKey(#[from] bdk_wallet::keys::KeyError),
+}
+
+impl
+    From<
+        bdk_wallet::LoadWithPersistError<
+            <crate::wallet::Persistence as bdk_wallet::AsyncWalletPersister>::Error,
+        >,
+    > for InitWalletFromMnemonic
+{
```
torkelrogstad, line 367: Ah, such a lovely type

lib/wallet/error.rs
```diff
@@ -332,9 +796,317 @@ where
...
+#[derive(Diagnostic, Debug, Error)]
+pub enum SendWalletTransaction {
+    #[error("failed to broadcast tx")]
+    BroadcastTx(#[source] jsonrpsee::core::ClientError),
+    #[error(transparent)]
+    CreateSendPsbt(#[from] CreateSendPsbt),
+    #[error(transparent)]
+    SignTransaction(#[from] WalletSignTransaction),
+    #[error("failed to broadcast send transaction (OP_DRIVECHAIN not supported by node)")]
+    OpDrivechainNotSupported,
```
torkelrogstad, line 1044: Could consider adding an explanation for how to fix this, similar to how we do with a few other obscure errors. This relates to the node not having `acceptnonstdtxn`, right?

lib/wallet/error.rs
```diff
@@ -144,6 +190,249 @@ pub enum WalletSync {
...
+}
+
+type Persistence = <crate::wallet::Persistence as bdk_wallet::AsyncWalletPersister>::Error;
+
+#[derive(Debug, Diagnostic, Error)]
+pub enum InitWalletFromMnemonic {
+    #[error("failed to create wallet")]
+    CreateWallet(#[from] bdk_wallet::CreateWithPersistError<Persistence>),
+    #[diagnostic(transparent)]
+    #[error("wallet data mismatch! Wipe your data directory and try again.")]
```
torkelrogstad, line 351: might be better if we have the error message as `wallet data mismatch (wipe your data directory and try again)`

Capital letters and dots compose poorly with nested errors

lib/wallet/error.rs
torkelrogstad: Okay, I see now that you don't actually include the plaintext mnemonic in the error later on. Wouldn't it be better to have a bool that indicates plaintext/encrypted?

---

Final realization: you're not even changing anything, just moving around. Feel free to ignore! Even though I think it's still a good idea to adjust this error

torkelrogstad (commented):
In general really nice stuff!

lib/server.rs
Ash-L2L: The downcasting approach is unreliable - if an error type gets wrapped in some other error type, the downcasting fails, and the status codes will not work as intended.
IMO it is best not to implement this for `miette::Report`, and to explicitly compose status codes

lib/util.rs
Ash-L2L: Moved to `lib/display.rs`

lib/wallet/error.rs
Ash-L2L: Fixed, now uses bool values to indicate if the values are `Some` or `None`

lib/wallet/error.rs
Ash-L2L: Fixed, using a type alias to clean this up now

torkelrogstad (approved):

PR #217, opened 2025-05-21
multi: implement file size based log rotation

torkelrogstad:
Log sizes can baloon in size - we've seen examples in the wild of files well over 1.5GB!

This relies on a long-standing, unmerged PR to tokio-rs/tracing: https://github.com/tokio-rs/tracing/pull/2497

The author has forked tokio-rs/tracing and applied the commits from the
PR to the most recent release: https://github.com/torkelrogstad/tokio-tracing/tree/log-rotation-size


Cargo.toml
```diff
@@ -68,3 +68,21 @@ let_underscore_drop = "deny"
 [patch.crates-io.temp-dir]
 git = "https://gitlab.com/A-Manning/leonhard-llc-ops.git"
 branch = "temp-dir-leak"
+
+# Have to use patches here. Depending directly on the fork breaks macros. 
+
+[patch.crates-io.tracing]
```
Ash-L2L, line 74: Why do we need all of these? Could you just change the dependency on `tracing-subscriber` to use crates.io instead of a path in your fork (https://github.com/torkelrogstad/tokio-tracing/blob/master/tracing-appender/Cargo.toml#L30)?

lib/cli.rs
```diff
@@ -63,11 +62,11 @@ fn get_data_dir() -> Result<PathBuf, String> {
 
 // Sub-par location for the log file.
 // https://github.com/LayerTwo-Labs/bip300301_enforcer/issues/133
-const DEFAULT_LOG_FILENAME: &str = "bip300301_enforcer.log";
+const LOG_FILENAME: &str = "bip300301_enforcer.log";
 
 // Sub-par location for the log dir.
 // https://github.com/LayerTwo-Labs/bip300301_enforcer/issues/133
-const DEFAULT_LOG_DIRNAME: &str = "logs";
+const LOG_DIRNAME: &str = "logs";
```
Ash-L2L, line 69: Prefer `DEFAULT_LOG_DIRNAME` here, since the user can set it to something else

Ash-L2L (changes requested):

Ash-L2L (approved):

PR #218, opened 2025-05-21
Fix stuck sync due to off-by-one error, simplify header sync

Ash-L2L:


.github/workflows/check_lint_build_release.yaml
```diff
@@ -125,7 +125,7 @@ jobs:
         with:
           # Stable, as of January 2015. If updating, also update
```
torkelrogstad, line 126: Should update the comment

lib/errors.rs
```diff
@@ -0,0 +1,83 @@
...
+        std::fmt::Display::fmt(&self.0, f)?;
+        let mut source: Option<&dyn std::error::Error> = self.0.source();
+        while let Some(cause) = source {
+            std::fmt::Display::fmt(": ", f)?;
+            std::fmt::Display::fmt(cause, f)?;
+            source = cause.source();
+        }
+        Ok(())
+    }
+}
```
torkelrogstad, line 31: Could you add a unit test for this? We really don't have much in the way of those, and the small parts of our codebase that is easily unit testable really should be

torkelrogstad: A benefit of unit tests for things like this is that it also serves as an example of what usage and results look like

.github/workflows/check_lint_build_release.yaml
torkelrogstad: What do you think of this? https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/220

Lets us simplify down to only one place we specify the Rust version. Also has the benefit of getting rustup users (AKA everyone?) to automatically be on the correct version

lib/validator/dbs/block_hashes.rs
```diff
@@ -184,6 +194,7 @@ impl BlockHashDbs {
     }
 
     /// Check if the database contains the provided block
+    #[allow(dead_code)]
```
torkelrogstad, line 197: There's little point in keeping this around if it's not used, IMO

lib/validator/dbs/block_hashes.rs
```diff
@@ -56,6 +56,16 @@ pub mod error {
...
     }
 
+    #[derive(Debug, Error)]
+    pub(crate) enum LastCommonAncestor {
+        #[error(transparent)]
+        DbGet(#[from] db_error::Get),
+        #[error(transparent)]
+        DbTryGet(#[from] db_error::TryGet),
+        #[error("missing or incorrect ancestor/height data for {0}")]
+        MissingData(BlockHash),
```
torkelrogstad, line 66: I think it'd be helpful to include the height here as well, for debuggability

lib/validator/dbs/block_hashes.rs
```diff
@@ -280,6 +291,60 @@ impl BlockHashDbs {
...
+            return Ok(block_hash0);
+        }
+        if block_hash0 == BlockHash::all_zeros() || block_hash1 == BlockHash::all_zeros() {
+            return Ok(BlockHash::all_zeros());
+        }
+        let height0 = self.height.get(rotxn, &block_hash0)?;
+        let height1 = self.height.get(rotxn, &block_hash1)?;
+        let mut ancestors0 = self.ancestor_headers(rotxn, block_hash0);
+        let mut ancestors1 = self.ancestor_headers(rotxn, block_hash1);
+        // Equalize heights to min(height0, height1)
```
torkelrogstad, line 314: This comment looks like it should be placed further down? What you're doing here appears to be verifying that both sets of ancestors have enough items?

torkelrogstad: Okay, scratch that. What you're doing is iterating on the tallest block hash ancestor set, in a way that they actually end up equalized. Not really immediately clear, a longer comment would be great

lib/validator/dbs/block_hashes.rs
```diff
@@ -280,6 +291,60 @@ impl BlockHashDbs {
...
+        }
+        for _ in 0..=std::cmp::min(height0, height1) {
+            let ancestor0 = ancestors0
+                .next()?
+                .ok_or_else(|| error::LastCommonAncestor::MissingData(block_hash0))?
+                .0;
+            let ancestor1 = ancestors1
+                .next()?
+                .ok_or_else(|| error::LastCommonAncestor::MissingData(block_hash1))?
+                .0;
```
torkelrogstad, line 340: nit: I think destructuring syntax is more readable than the `.0`. `let (ancestor1, _) = ancestor1.next().ok_or_else(..)?`

torkelrogstad (commented):
Only minor comments, really good stuff

torkelrogstad (commented):
LGTM 👌

PR #222, opened 2025-05-30
wallet: apply unconfirmed TXs to the wallet post broadcast

### LINKED ISSUE #161: wallet: isn't synced immediately after a transaction
When I send a transaction then refetch the transaction list, the latest transaction isn't included as an unconfirmed transaction.

I suspect the wallet isn't synced immediately after the transaction is sent. I expect the transaction to be immediately available after I get a txid from the Send-endpoint. Ditto for all other actions that move funds to/from the wallet.

torkelrogstad:
Closes #161

### DIFF
--- a/lib/rpc_client.rs
+++ b/lib/rpc_client.rs
@@ -65,7 +65,7 @@ pub fn create_client(
 }
 
 /// Broadcasts a transaction to the Bitcoin network.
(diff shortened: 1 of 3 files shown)


A-Manning (approved):
LGTM

Issue #207, opened 2025-05-05
slow wallet sync on startup


octobocto:
### The problem
On my machine, it takes ~13-23 seconds for the enforcer wallet to sync. In the meantime, I can't fetch transactions, utxos or balance.

### Desired behavior
Be fast! The users of bitwindow want to see their wallet-stuff the second the enforcer is booted.

### More info/logs
```
flutter: level=debug msg="/Users/bo/code/drivechain-frontends/bitwindow/assets/bin/bip300301-enforcer: 2025-05-05T12:17:49.614562Z DEBUG sync_to_tip: bip300301_enforcer_lib::wallet::cusf_block_producer: lib/wallet/cusf_block_producer.rs:53: Synced validator, syncing wallet.. tip_hash=00000292efd1fe2967d2ed540444d7ffa132f2e6d062bc853468b2cde7148f05
flutter: level=debug msg="/Users/bo/code/drivechain-frontends/bitwindow/assets/bin/bip300301-enforcer: 2025-05-05T12:17:49.614784Z DEBUG sync_to_tip: bip300301_enforcer_lib::wallet::cusf_block_producer: lib/wallet/cusf_block_producer.rs:75: Sending start signal to wallet task, kicking off periodic sync
wait 23 seconds...
flutter: level=debug msg="/Users/bo/code/drivechain-frontends/bitwindow/assets/bin/bip300301-enforcer: 2025-05-05T12:18:02.371700Z DEBUG grpc_server:get_utxos: bip300301_enforcer_lib::wallet: lib/wallet/mod.rs:1630: listed 16 wallet utxos in 1.497833ms method=POST uri=http://localhost:50051/cusf.mainchain.v1.WalletService/ListUnspentOutputs request_id="req_c12a671951e64fa7b56d11a65a01c8e3"
```

I currently have 22 transactions in the wallet.

torkelrogstad:
There's actually several things going on here at the same time which it could potentially be a good idea to adjust. 

1. Wallet sync source. By default, the wallet syncs agains Electrum. I'm pretty sure that syncing against Esplora is faster. This can be configured, or we can just update the default. 
2. We're waiting with kicking off the sync until the validator is synced. There's actually no good reason to do it this way, I believe. When doing the sync the external source is treated as the source of truth about the chain, anyways
[...]

PR #230, opened 2025-06-17
multi: refactor task structure, add shutdown to JSON-RPC server

torkelrogstad:
1. Instead of implicitly spawning a wallet sync task on wallet instantation, make this explicit and spawn from main.rs. This allows us to rip out some start/stop complexity, and also allows us to spawn the wallet sync task right away, instead of waiting until the validator has synced.

2. Remove callback usage from the mempool sync task. Instead return the synced mempool, and allow what used to be a callback instead be a regular function call on the mempool. Allows to reduce some complexity around senders/receivers.

3. Implement support for shutting down the `getblocktemplate` JSON-RPC server. Previously this was never cancelled, leading us to always hit the force shutdown after a short timeout.

4. Implement support for shutting down the main validator JSON-RPC server

This fixes the second point of this comment: https://github.com/LayerTwo-Labs/bip300301_enforcer/issues/207#issuecomment-2922748187

### DIFF
--- a/app/main.rs
+++ b/app/main.rs
@@ -2,7 +2,7 @@ use std::{future::Future, net::SocketAddr, time::Duration};
 
 use bdk_wallet::bip39::{Language, Mnemonic};
 use bip300301_enforcer_lib::{
-    cli::{self, LogFormatter},
+    cli::{self, LogFormatter, WalletSyncSource},
     errors::ErrorChain,
     p2p::compute_signet_magic,
     proto::{
@@ -420,14 +420,14 @@ async fn spawn_gbt_server(
     Ok(handle)
 }
 
-async fn run_gbt_server(
+async fn start_gbt_server(
     mining_reward_address: bitcoin::Address,
     network: bitcoin::Network,
     network_info: bitcoin_jsonrpsee::client::NetworkInfo,
     sample_block_template: bitcoin_jsonrpsee::client::BlockTemplate,
     mempool: cusf_enforcer_mempool::mempool::MempoolSync<Wallet>,
     serve_addr: SocketAddr,
-) -> miette::Result<()> {
+) -> miette::Result<jsonrpsee::server::ServerHandle> {
     let gbt_server = cusf_enforcer_mempool::server::Server::new(
         mining_reward_address.script_pubkey(),
         mempool,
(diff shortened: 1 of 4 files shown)


lib/wallet/mod.rs
```diff
@@ -787,12 +685,46 @@ impl Wallet {
...
-        Ok(Self {
-            inner,
-            task: Arc::new(task),
-            has_initial_synced: Arc::new(AtomicBool::new(false)),
-        })
+        Ok(Self { inner })
+    }
+
+    pub async fn sync_task<F: Future<Output = ()>>(
+        &self,
```
Ash-L2L, line 692: There should only ever be one sync task per wallet, and it should always start when calling `Wallet::new(...)`.

IMO it is best to either
* keep the task as part of the `Wallet` struct
* return a future from `Wallet::new(...)`, ie. `pub async fn new(...) -> Result<(Self, TaskFuture), error::InitWallet>`

app/main.rs
```diff
@@ -1008,38 +1016,120 @@ async fn main() -> Result<()> {
...
+            let mut task_labels = vec!["main task", "JSON-RPC server"];
+
+            if let Some(wallet_sync_task) = wallet_sync_task {
+                tasks.push(Box::pin(
+                    wallet_sync_task.inspect(|_| tracing::info!("wallet sync task finished")),
+                ));
+                task_labels.push("wallet sync task");
+            }
+
+            // Is there a better way to do this?
```
Ash-L2L, line 1097: You probably want to use `futures::try_join` or `tokio::try_join`.
The function should probably be renamed to `TaskHandles::try_join_all` if you don't want to wait for all tasks to finish, if one of them completes with an error.

Alternatively, you could map the `tasks` array like
```rust
let mut tasks: Vec<std::pin::Pin<Box<dyn Future<Output = (&'static str, Result<_, _>)> + Send>>> = vec![
    Box::pin(main_task.map(|res| {
        tracing::info!("main task finished");
        ("main task", res)
    })),
    Box::pin(json_rpc_handle.map(|res| {
        tracing::info!("JSON-RPC server finished");
        ("JSON-RPC server", res)
    })),
];
```
and then

```rust
for (task_name, result) in futures::future::join_all(tasks).await.into_iter()  { ... }
```

Ash-L2L (changes requested):

PR #258, opened 2025-09-10
integration tests: dump core+enforcer logs on failure

torkelrogstad:
Right now the integration test failures are pretty inscrutable, especially on CI. This PR attempts to remedy that by dumping file paths for Enforcer + Core logs, as well as the last parts of the files.

### DIFF
--- a/integration_tests/test_peer_bmm_request.rs
+++ b/integration_tests/test_peer_bmm_request.rs
@@ -1,13 +1,11 @@
 use bip300301_enforcer_lib::{bins::CommandExt, proto::common::ConsensusHex};
-use futures::{StreamExt, channel::mpsc};
+use futures::channel::mpsc;
 use tokio::time::sleep;
-use tracing::Instrument;
 
 use crate::{
     integration_test::{activate_sidechain, fund_enforcer, propose_sidechain},
     mine,
-    setup::{DummySidechain, Mode, Network, PostSetup, Sidechain, setup},
-    util::{self, BinPaths},
+    setup::{DummySidechain, PostSetup, Sidechain},
 };
 
 async fn create_bmm_request_tx(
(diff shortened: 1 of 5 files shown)


integration_tests/util.rs
```diff
@@ -81,24 +169,116 @@ impl<Fut> AsyncTrial<Fut> {
...
         let span = tracing::info_span!("test", name = %self.name);
+        let test_name = self.name.clone();
+        let file_registry = self.file_registry.clone();
+
         libtest_mimic::Trial::test(self.name, move || {
             span.in_scope(|| {
-                rt_handle.block_on(async { self.test.map_err(libtest_mimic::Failed::from).await })
+                rt_handle.block_on(async {
+                    let result = self.test.map_err(libtest_mimic::Failed::from).await;
+                    result.inspect_err(|_err| {
```
Ash-L2L, line 195: If you're not using the error, you can just do `if result.is_err() && let Some(registry) = &file_registry { ... }`

Ash-L2L, line 195: I'm not sure if this is the right place to display logs on test failure, since other tests will be running concurrently, and the output might get mixed with output from other tests.

Is it possible to collect these and display the logs for failing tests sequentially, after the test harness runs in `main.rs`?

integration_tests/test_peer_bmm_request.rs
```diff
@@ -130,23 +116,6 @@ async fn test_peer_bmm_request_task(
...
-        }
-        .in_current_span()
-    })
-    .into();
-    res_rx
-        .next()
-        .await
-        .ok_or_else(|| anyhow::anyhow!("Unexpected end of test task result stream"))?
+pub async fn test_peer_bmm_request(mut post_setup: PostSetup) -> anyhow::Result<()> {
+    test_peer_bmm_request_task(&mut post_setup, mpsc::unbounded().0).await
```
Ash-L2L, line 120: I don't think this is equivalent. `res_rx` is used to receive errors from the test here, both from background tasks and from the actual test. This would not capture any errors from background tasks

integration_tests/util.rs
```diff
@@ -68,9 +70,95 @@ impl BinPaths {
...
+        };
+        registry
+            .entry(test_name.to_string())
+            .or_default()
+            .push(file_with_config);
+    }
+
+    /// Get all registered files with their configurations for a test
+    fn get_files(&self, test_name: &str) -> Vec<FileWithConfig> {
+        let registry = self.inner.lock().unwrap();
```
Ash-L2L, line 153: You can use `parking_lot`'s mutex if you don't want to handle poisoning

Ash-L2L (approved):

PR #334, opened 2026-05-04
validator: report M8 without matching M7 as fatal

torkelrogstad:
BIP 301 says a block containing an M8 without the corresponding M7 MUST be considered invalid. The enforcer was treating this as non-fatal. Rename `NotAcceptedByMiners` to `NoCorrespondingM7` to match spec wording.

### DIFF
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -462,9 +462,10 @@ fn handle_m5_m6(
     }
 }
 
-/// Handles a (potential) M8 BMM request.
-/// Returns `true` if this is a valid BMM request, `HandleM8Error::Jfyi` if
-/// this is an invalid BMM request, and `false` if this is not a BMM request.
+/// Validate a (potential) M8 BMM request.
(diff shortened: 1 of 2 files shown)


lib/validator/task/mod.rs
```diff
@@ -1541,22 +1542,35 @@ mod tests {
...
         let err =
             handle_m8(&tx, Some(&wrong), &prev_hash).expect_err("wrong commitment must reject");
-        assert!(matches!(err, error::HandleM8::NotAcceptedByMiners));
+        assert!(matches!(err, error::HandleM8::NoCorrespondingM7));
+        assert!(err.is_fatal());
+    }
+
+    #[test]
+    fn handle_m8_expired_request_is_non_fatal() {
+        // The spec does not mandate any behavior when M8.prev_mainchain_block_hash
```
Ash-L2L, line 1568: Could you add this comment to the `HandleM8::BmmRequestExpired` error variant?

Ash-L2L (approved):

Ash-L2L (changes requested):
This should not be a fatal error - we don't want to crash the enforcer on this error, only reject a tx

PR #336, opened 2026-05-04
validator: mark M6 errors as fatal

torkelrogstad:


Ash-L2L (approved):

Ash-L2L (changes requested):
These should not be a fatal errors - we don't want to crash the enforcer on these errors, only reject txs

PR #338, opened 2026-05-06
Declare all dependencies in `[workspace.dependencies]`, fatality -> error_fatality

Ash-L2L:


torkelrogstad (approved):
Replaces a bad version of a library with one we maintain ourselves.

PR #347, opened 2026-05-18
app: refactor task management to avoid races

torkelrogstad:
Prior to this commit we would get `Unable to receive error from gRPC server: oneshot canceled` every now and then for errors that happened elsewhere. This made it really hard to understand what was actually erroring.

### DIFF
--- a/app/main.rs
+++ b/app/main.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, net::SocketAddr, time::Duration};
+use std::{net::SocketAddr, time::Duration};
 
 use bdk_wallet::bip39::{Language, Mnemonic};
 use bip300301_enforcer_lib::{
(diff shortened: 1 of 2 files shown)


app/main.rs
```diff
@@ -374,14 +346,22 @@ async fn is_address_port_open(addr: &str) -> Result<bool, std::io::Error> {
...
 
+/// Initialize a mempool sync. Returns the sync handle plus a receiver that
+/// fires if the background sync task hits an error. Caller should
+/// race the receiver against the shutdown signal
 async fn sync_mempool<Enforcer, RpcClient>(
     mut enforcer: Enforcer,
     network: bitcoin::Network,
     rpc_client: RpcClient,
     zmq_addr_sequence: &str,
-    err_tx: oneshot::Sender<error::MempoolTask<Enforcer>>,
```
Ash-L2L, line 382: Typically, it is preferable for the user to provide a channel, rather than returning one - the user has more flexibility this way.
In this case however, I think it is fine to return a receiver

Ash-L2L (approved):

PR #353, opened 2026-05-21
validator: verify required M6 IDs, simplify integration tests

torkelrogstad:
Implement verification of presence of withdrawal bundles. If they exceed
the vote count threshold they MUST be included in the
corresponding block. This also leads to implementing corresponding fixes
in our mining logic, ensuring we don't produce invalid blocks. 

Also refactor and simplify integration tests somewhat

TODO: 
- [ ] fix block invalidation in cusf-enforcer-mempool when running in mempool mode

### DIFF
--- a/lib/wallet/cusf_block_producer.rs
+++ b/lib/wallet/cusf_block_producer.rs
@@ -252,6 +252,9 @@ impl CusfEnforcer for Wallet {
     }
 }
 
+// TODO: make ACKing configurable for the miner
+const ACK_ALL_PROPOSALS: bool = true;
+
 impl CusfBlockProducer for Wallet {
     type InitialBlockTemplateError = error::InitialBlockTemplate;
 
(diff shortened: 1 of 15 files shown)


lib/wallet/cusf_block_producer.rs
```diff
@@ -328,6 +330,17 @@ impl CusfBlockProducer for Wallet {
...
+    /// 1. RPC server (within the `cusf_enforcer_mempool` crate) calls
+    ///    `initial_block_template` (see above) for the coinbase outputs
+    /// 2. Fills in `prefix_txs` with the mempool txs it wants to include
+    /// 3. Calls this function for the trailing bits (this function!)
+    /// 4. Stitches it all together and spits out a block template to the client
+    ///
+    /// This function is our "hook" for adding M6 withdrawal txs to the end of
+    /// the block, and tacking the matching M7 BMM-accept outputs onto the
+    /// coinbase for any M8s pulled in from the mempool.
     async fn block_template_suffix<const COINBASE_TXN: bool>(
```
Ash-L2L, line 344: This  function should verify that the corresponding M6 tx exists if an M4 ack would cause a proposed withdrawal to reach the withdrawal threshold.

Ash-L2L (changes requested):

PR #357, opened 2026-05-27
multi: switch to checked-in generated protobuf/grpc code

torkelrogstad:
Avoids mucking with Git sub modules, and trades away some build dependencies in favor of a big diff of generated code.


lib/proto.rs
```diff
@@ -212,775 +216,20 @@ impl From<Error> for tonic::Status {
...
-    }
-
-    impl From<HeaderSyncProgress> for SubscribeHeaderSyncProgressResponse {
-        fn from(progress: HeaderSyncProgress) -> Self {
-            Self {
-                current_height: progress.current_height,
-            }
-        }
-    }
+    pub use super::cusf::mainchain::v1::*;
```
Ash-L2L: Is it possible to remove the wildcard exports?

eg.
```rust
pub use super::cusf::{
    mainchain::v1 as mainchain,
    sidechain::v1 as sidechain,
};
```

Ash-L2L (changes requested):

lib/proto/cusf/crypto.rs
```diff
@@ -0,0 +1,7 @@
+#[expect(clippy::allow_attributes, reason = "generated code uses #[allow]")]
+#[path = "../generated/cusf.crypto.v1.rs"]
+mod generated;
+
+pub mod v1 {
```
Ash-L2L: This could be
`pub use generated as v1`;

lib/proto/cusf/sidechain.rs
```diff
@@ -0,0 +1,7 @@
+#[expect(clippy::allow_attributes, reason = "generated code uses #[allow]")]
+#[path = "../generated/cusf.sidechain.v1.rs"]
+mod generated;
+
+pub mod v1 {
```
Ash-L2L: This could be
`pub use generated as v1;`

Ash-L2L (changes requested):

Ash-L2L (approved):

PR #362, opened 2026-05-29
Fix op drivechain spends

Ash-L2L:


lib/validator/task/mod.rs
```diff
@@ -70,7 +70,8 @@ impl<'a> BlockHandler<'a> {
 const LEADING_BY_50_MARGIN: u16 = 50;
 
 /// Deposit or (sidechain_id, m6id, sequence_number)
```
torkelrogstad: Comment should be updated

lib/validator/task/mod.rs
```diff
@@ -604,107 +600,149 @@ impl BlockHandler<'_> {
         &self,
         rotxn: &RoTxn,
```
torkelrogstad, line 605: Ah

lib/validator/task/mod.rs
```diff
@@ -604,107 +600,149 @@ impl BlockHandler<'_> {
...
+                            {
+                                error::HandleM5M6::InvalidM6
+                            }
+                        });
+                    } else {
+                        res = Some(Either::Right((
+                            (sidechain_number, m6id, sequence_number),
+                            diff,
+                        )));
+                    }
```
torkelrogstad: This reads strange IMO. `if let` deconstruction + mutation + early return?

torkelrogstad: Another thing: the for loop body is very big, making it a bit hard to at first glance see exactly what `res` is here. Would it be an idea to do a standalone function for the loop body?

torkelrogstad (commented):
Would be great with an integration test asserting on an invalid block, IMO

lib/validator/task/mod.rs
Ash-L2L: changed the `if let ...` block to a single match for readability, and introduced an enum to make the `res` cases more explicit

torkelrogstad (approved):

PR #369, opened 2026-06-04
Reject transactions that spend a treasury UTXO without creating a new one

1440000bytes:
Per BIP300 a transaction that spends a treasury UTXO as an input and creates no new treasury UTXO as an output is invalid.

https://github.com/LayerTwo-Labs/bip300_bip301_specifications/blob/master/bip300.md#validation-rules-specification

`handle_m5_m6` builds `ctip_spends` (treasury inputs) but only consults it inside the loop over `new_ctips` (treasury outputs). A tx that spends a treasury UTXO and creates no new treasury output never enters that loop, so it returns `Ok(None)` and is ignored. OP_DRIVECHAIN is anyone-can-spend, so anyone could drain a sidechain's treasury this way.

This adds the missing check. Every spent treasury must have a corresponding new treasury output, otherwise the tx is invalid. Block-level rejection of the invalid tx is handled by #365 (which propagates non-fatal tx errors instead of
skipping them).

### DIFF
--- a/lib/validator/task/error.rs
+++ b/lib/validator/task/error.rs
@@ -243,6 +243,15 @@ pub(in crate::validator) enum HandleM5M6 {
     #[error("Old Ctip for sidechain {} is unspent", .sidechain_number.0)]
     #[fatal(false)]
     OldCtipUnspent { sidechain_number: SidechainNumber },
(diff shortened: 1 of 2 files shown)


Ash-L2L (approved):

Ash-L2L:
Thanks! LGTM

PR #370, opened 2026-06-04
Reject re-proposing an already-pending withdrawal bundle (M3)

1440000bytes:
Per BIP300 re-proposing an M6ID that was already proposed in a previous block and not yet paid out must invalidate the block.

`handle_m3_propose_bundle` only checked that the slot is active, and `put_pending_m6id` overwrites an existing entry with a fresh `PendingM6idInfo` (vote_count reset to 1, proposal_height reset). So a miner could reset any pending withdrawal bundle's ack count and age by re-proposing its M6ID via M3, indefinitely blocking a withdrawal or reviving an expired bundle. The M1 path already has this anti-reset protection; M3 did not.

## Fix

Reject an M3 that re-proposes an already-pending M6ID. Re-proposing an expired or paid-out bundle (no longer pending) is still allowed.

### DIFF
--- a/lib/validator/task/error.rs
+++ b/lib/validator/task/error.rs
@@ -62,9 +62,21 @@ impl From<db::Error> for HandleFailedSidechainProposals {
 }
 
 #[derive(Debug, Error, Fatality, Split, Transitive)]
+#[expect(clippy::duplicated_attributes)]
 #[split(attrs(derive(Debug, Error)))]
-#[transitive(from(db::error::TryGet, db::Error))]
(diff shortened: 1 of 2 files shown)


torkelrogstad (approved):
Thank you! I see that there's actually a slight mismatch between the spec and the code here wrt. expired bundles. I made a clarifying PR to the spec to remedy that: https://github.com/LayerTwo-Labs/bip300_bip301_specifications/pull/5

Ash-L2L (approved):

PR #373, opened 2026-06-04
Handle zero-input blinded M6 transactions

ekulkisnek:
## Summary
- accept zero-input blinded M6 transactions in BroadcastWithdrawalBundle without mis-parsing them as segwit marker/flag bytes
- store zero-input blinded M6 bundle proposals with an explicit input-count VarInt instead of witness marker encoding
- reload stored bundle proposals with the same zero-input fallback parser so M3 proposal mining works after restart

## Verification
- cargo check -p bip300301_enforcer_lib --lib
- local Liquid sidechain peg-out smoke: BroadcastWithdrawalBundle accepted M6id 12f164c44356a63417364a9fa532335aad14ace2745c90501dfeddd045350bb2; native enforcer template included M3; mined mainchain block 15281 reported sidechain 5 withdrawal bundle event submitted; next block 15282 included M4 ACK output.

### DIFF
--- a/lib/server/wallet/grpc.rs
+++ b/lib/server/wallet/grpc.rs
@@ -1,10 +1,14 @@
-use std::{borrow::Cow, collections::HashMap, str::FromStr, sync::Arc};
+use std::{borrow::Cow, collections::HashMap, io::Cursor, str::FromStr, sync::Arc};
 
 use bdk_wallet::bip39::Mnemonic;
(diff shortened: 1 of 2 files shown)


ekulkisnek:
this is for allowing the M6 transaction to show up with elements confidential addresses on a liquid sidechain.

torkelrogstad:
Thanks for implementing this! I've taken your fixes and expanded on them in #385. I've added you as a git commit co-author.

PR #384, opened 2026-06-10
wallet: add BIP47 and silent payments support, regenerate protos

mplatt8:


lib/wallet/mod.rs
```diff
@@ -2304,4 +2378,801 @@ impl Wallet {
...
+        self.inner.ensure_scanner_activated().await?;
+        let connection = self.inner.self_db.lock().await;
+        let next_m: i64 = connection
+            .query_row(
+                "SELECT COALESCE(MAX(m), 0) + 1 FROM silent_payment_labels WHERE m >= 1",
+                [],
+                |row| row.get(0),
+            )
+            .map_err(error::ReusablePayments::db)?;
+        let m: u32 = u32::try_from(next_m).unwrap_or(1);
```
torkelrogstad, line 2414: Why aren't we just reading out an u32 from the DB? Hiding an error by unwrapping to 1 seems wrong

lib/wallet/mod.rs
```diff
@@ -2304,4 +2378,801 @@ impl Wallet {
...
+                 ORDER BY p.first_seen_unix ASC",
+            )
+            .map_err(error::ReusablePayments::db)?;
+        let rows_result: Result<Vec<(String, u32, i64, i64)>, _> = stmt
+            .query_map([], |row| {
+                Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?))
+            })
+            .and_then(|iter| iter.collect());
+        drop(stmt);
+        drop(connection);
```
torkelrogstad, line 2476: We can avoid manual drops by instead processing this within a `{}` block

lib/wallet/error.rs
```diff
@@ -1406,3 +1406,65 @@ impl ToStatus for GetNewAddress {
         }
     }
 }
+
+#[derive(Debug, Diagnostic, Error)]
+pub enum ReusablePayments {
```
torkelrogstad, line 1411: We should make sure to not repeat errors here and other places. You've reused `NotUnlocked`, should do the same for the other repeated as well

lib/wallet/mod.rs
```diff
@@ -117,6 +118,9 @@ struct WalletInner {
     config: Config,
     /// Limits GenerateBlocks to one concurrent call at a time.
     generate_blocks_semaphore: Arc<tokio::sync::Semaphore>,
+    bip47_send_locks:
+        tokio::sync::Mutex<std::collections::HashMap<String, Arc<tokio::sync::Mutex<()>>>>,
```
Ash-L2L: These locks should be semaphores. Using a mutex directly around something that is never mutated seems odd

Ash-L2L (changes requested):
Could the BIP47 impl be moved into a standalone crate?

lib/wallet/reusable_payments/impl.rs
```diff
@@ -0,0 +1,1122 @@
...
+
+use super::{
+    bip47,
+    scan::{
+        ScanContext, ScanCursor, ScanEvent, ScannedInput, ScannedTx, ScrubOnDrop,
+        extract_bip47_designated_pubkey, extract_input_pubkey, scan_tx,
+    },
+    silent_payments, util,
+};
+use crate::wallet::{WalletInner, error};
```
Ash-L2L, line 24: Nit: these can all be part of a `use crate::{...};` statement. You don't need both `use super` and `use crate`.

lib/wallet/reusable_payments/mod.rs
```diff
@@ -0,0 +1,42 @@
+pub use ::bip47;
```
Ash-L2L, line 1: Why is this re-exported?

You don't need to import from root here either, should be able to do

```rust
use bip47::{PaymentCode, Version as Bip47Version};

pub mod ...;
```

lib/wallet/reusable_payments/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+pub mod r#impl;
+pub mod scan;
+pub mod silent_payments;
+#[cfg(test)]
+mod silent_payments_test_vectors;
+mod util;
+
+pub use self::{
+    bip47::{PaymentCode, Version as Bip47Version},
```
Ash-L2L, line 10: Why are these re-exported?

lib/server/wallet/grpc.rs
```diff
@@ -852,6 +852,301 @@ impl WalletService for crate::wallet::Wallet {
...
+        Ok(tonic::Response::new(
+            crate::proto::mainchain::ListSilentPaymentReceivesResponse {
+                items,
+                scan_tip_height: tip_height,
+            },
+        ))
+    }
+}
+
+fn bip47_version_from_proto(
```
Ash-L2L, line 1118: Nit: please try to keep files readable top-to-bottom. These functions should appear before any uses, if possible

Ash-L2L (changes requested):

PR #390, opened 2026-06-12
multi: handle intra-block conflicting TXs in block generation

torkelrogstad:


Ash-L2L:
I think it would be better to use the `conflicts_with` field in [`cusf_enforcer_mempool::cusf_enforcer::TxAcceptAction`](https://github.com/LayerTwo-Labs/cusf-enforcer-mempool/blob/master/lib/cusf_enforcer.rs#L42) to declare a conflict between mempool txs that create a deposit for the same sidechain. That should be sufficient to prevent this bug

PR #391, opened 2026-06-15
[DoS] Fix 2 remotely-triggerable panics parsing M1 sidechain descriptions

1440000bytes:
I found these bugs using differential fuzzing. Like rust-bitcoin, the enforcer's custom parsers must never panic when processing adversarial input. A libFuzzer [harness](https://github.com/LayerTwo-Labs/bip300301_enforcer/commit/c1ff347a2c3389896438c40da47856df3de83d58) for the M1 description path reproduces both panics (e.g. the empty input). 

---

`SidechainDeclaration::try_parse` (`lib/types.rs`) panics on two classes of malformed input. Both are reachable from a miner-controlled M1 sidechain proposal, so a single crafted coinbase can crash any enforcer that later reads
the stored proposal.

An M1 proposal's `description` is parsed from the coinbase with `rest` (`M1ProposeSidechain::parse`), so it is arbitrary, attacker-controlled, and may be empty or shorter than the fixed trailing fields. `handle_m1_propose_sidechain` stores it **without** parsing it as a declaration, so the malformed bytes are
accepted into state and persist.

The declaration is only parsed later, in two places, both of which panic rather than erroring:

- `Validator::sync_state_summary` runs automatically in `exit-after-sync` / state-digest mode (`app/main.rs`),
  so the node crashes while computing its consensus state summary. No client interaction required.
  
- The validator gRPC sidechain query crashes the node on demand for any caller. Note both call sites use `.ok()`, which discards an `Err` but does **not** stop a panic from unwinding.

## Bugs and fixes

1. **Out-of-bounds index on empty description.** When the description is empty, the version `tag` fails and the error closure built `UnknownVersion(input[0])`, indexing an empty slice. Fixed by reporting the version only when a byte is present, and surfacing the underlying parse error otherwise.

2. **Integer underflow on short description.** The description length is `remaining input - 52` (the trailing 32-byte and 20-byte hash fields). With fewer than 52 bytes left after the title, the `usize` subtraction underflows (panics under overflow checks; produces a wild `take` length otherwise). Fixed with `checked_sub`, failing on malformed input.

## Tests

Adds two regression tests (`test_try_deserialize_empty_description`,`test_try_deserialize_too_short_for_hashes`).

### DIFF
--- a/lib/types.rs
+++ b/lib/types.rs
@@ -328,8 +328,14 @@ impl SidechainDeclaration {
         }
         const VERSION_0: u8 = 0;
         let (input, _) = tag([VERSION_0].as_slice())(input).map_err(
-            |_: nom::Err<nom::error::Error<&[u8]>>| {
-                nom::Err::Error(ParseSidechainDeclarationError::UnknownVersion(input[0]))
+            |err: nom::Err<nom::error::Error<&[u8]>>| match input.first() {
+                // The version byte is present but unrecognized.
+                Some(&version) => {
+                    nom::Err::Error(ParseSidechainDeclarationError::UnknownVersion(version))
+                }
+                // Empty input: there is no version byte to report. Surface the
+                // underlying `tag` failure instead of indexing out of bounds.
+                None => failed_to_deserialize(err),
             },
         )?;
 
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
Thank you!

Ash-L2L (approved):

PR #392, opened 2026-06-15
[validation] fix sidechain proposal-age underflow in M2 ack

1440000bytes:
Found this bug with fuzzing of `connect_block`/`disconnect_block` (reorg round-trip) with fuzzer-driven block heights and seeded proposals.

`handle_m2_ack_sidechain` computes the proposal age with an unguarded `u32` subtraction:

```rust
let sidechain_proposal_age = height - sidechain.status.proposal_height;
```

If the enforcer holds a sidechain proposal whose `proposal_height` is greater than the height at which the M2 ack is processed, this underflows. The sibling `handle_failed_sidechain_proposals` already performs the **same** computation with `saturating_sub` and an explicit comment that this case happens:

```rust
// doing `height - sidechain.status.proposal_height` can panic if the enforcer has data
// from a previous sync that is not in the active chain.
let age = height.saturating_sub(sidechain.status.proposal_height);
```

`handle_m2_ack_sidechain` simply misses the same guard.

## Impact

- With overflow checks (tests/CI): panic while validating a block carrying the M2 ack - a node crash on a block.
- In a release build (overflow checks off): the subtraction wraps to a huge age, so `age <= max_age` is false and the sidechain silently fails to activate and two nodes with different proposal-data histories can disagree on activation.

The triggering condition (a proposal with `proposal_height > height`) is the same "data from a previous sync that is not in the active chain" scenario the sibling function documents and guards against.

## Fix

Use `height.saturating_sub(sidechain.status.proposal_height)`, matching `handle_failed_sidechain_proposals`.

## Test

Adds `handle_m2_ack_below_proposal_height_does_not_underflow`: a proposal recorded at height 100, acked at height 0, must not underflow.

### DIFF
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -2423,6 +2423,36 @@ mod tests {
 
     // ── handle_m2 ──
 
+    /// Regression: an M2 ack processed at a height below the stored proposal
+    /// height (e.g. a proposal from a previous sync that is not on the active
+    /// chain) must not underflow `height - proposal_height`. The sibling
+    /// `handle_failed_sidechain_proposals` already guards this case.
+    #[test]
+    fn handle_m2_ack_below_proposal_height_does_not_underflow() -> Result<()> {
+        let (_dir, dbs) = create_test_dbs()?;
+        let mut rwtxn = dbs.write_txn().into_diagnostic()?;
+        let handler = test_handler(&dbs);
+
+        let sidechain = test_sidechain(1, 100);
+        let proposal_id = sidechain.proposal.compute_id();
+        dbs.proposal_id_to_sidechain
+            .put(&mut rwtxn, &proposal_id, &sidechain)
+            .into_diagnostic()?;
+
+        // height 0 < proposal_height 100 must not panic / error.
+        let res = handler.handle_m2_ack_sidechain(
+            &rwtxn,
+            0,
+            SidechainNumber(1),
+            proposal_id.description_hash,
+        );
+        assert!(
+            res.is_ok(),
+            "M2 ack below proposal height must not underflow"
+        );
+        Ok(())
+    }
+
     #[test]
     fn handle_m2_activation_thresholds() -> Result<()> {
         let (_dir, dbs) = create_test_dbs()?;
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -222,7 +222,13 @@ impl BlockHandler<'_> {
             activated: false,
         };
 
-        let sidechain_proposal_age = height - sidechain.status.proposal_height;
+        // `height - proposal_height` can underflow if the enforcer holds a
+        // proposal from a previous sync that is not on the active chain (the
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
It is not clear to me how this could be reached in prod. But a saturating sub would in any case be safer.

1440000bytes:
```rust
#[test]
fn poc_connect_block_m2_ack_underflows_below_proposal_height() -> Result<()> {
    let (_dir, dbs) = create_test_dbs()?;
    let mut rwtxn = dbs.write_txn().into_diagnostic()?;

    // Prior-sync state: proposal recorded at height 100, one ack short of
    // the activation threshold, still pending in the proposal DB.
    let proposal_height = 100u32;
    let mut sidechain = test_sidechain(1, proposal_height);
    sidechain.status.vote_count = test_handler(&dbs)
        .thresholds
        .unused_sidechain_slot_activation_threshold;
    let proposal_id = sidechain.proposal.compute_id();
    dbs.proposal_id_to_sidechain
        .put(&mut rwtxn, &proposal_id, &sidechain)
        .into_diagnostic()?;

    // A real block at height 5 (< proposal_height) whose coinbase carries
    // an M2 ack for that proposal.
    let connect_height = 5u32;
    let m2_script: ScriptBuf = M2AckSidechain {
        sidechain_number: proposal_id.sidechain_number,
        description_hash: proposal_id.description_hash,
    }
    .try_into()
    .into_diagnostic()?;
    let block = build_test_block(
        BlockHash::all_zeros(),
        TestBlockParts {
            extra_coinbase_outputs: vec![TxOut {
                script_pubkey: m2_script,
                value: Amount::ZERO,
            }],
            ..Default::default()
        },
    );
    dbs.block_hashes
        .put_headers(&mut rwtxn, &[(block.header, connect_height)])
        .into_diagnostic()?;

    // Unguarded code panics here in any overflow-
    // checked build; with saturating_sub the age clamps to 0 and the ack
    // activates the sidechain.
    test_handler(&dbs)
        .connect_block(&mut rwtxn, &block)
        .into_diagnostic()?;

    let active = dbs
        .active_sidechains
        .sidechain()
        .try_get(&rwtxn, &proposal_id.sidechain_number)
        .into_diagnostic()?;
    assert!(
        active.is_some(),
        "ack at age 0 must activate the sidechain; a wrapping subtraction \
         would compute a huge age and silently drop the activation"
    );
    Ok(())
}
```

1. overflow-checked build (the cargo test / CI default, and any node built with -C overflow-checks). The node aborts mid-block while validating a valid block.:

```
thread '...poc_connect_block_m2_ack_underflows_below_proposal_height' panicked at
lib/validator/task/mod.rs:
attempt to subtract with overflow
```

2. --release (overflow-checks off): no panic, but the age wraps to ~2³², so age <= max_age is false and the activation is silently dropped — the active.is_some() assertion fails. A peer without the stale proposal would activate the same block - consensus divergence.

Ash-L2L (approved):

PR #393, opened 2026-06-16
[validation] fix reorg deleting a previously-active sidechain on activation undo

1440000bytes:
Found this bug with fuzzing of `connect_block` then `disconnect_block` (reorg round-trip), asserting the consensus state is exactly restored.

When an M2 ack activates a sidechain proposal into a slot that is **already occupied** by an active sidechain, the activation overwrites the incumbent — but a block disconnect (reorg) then *deletes* the slot instead of restoring it, permanently losing the previously-active sidechain.

`handle_m2_ack_sidechain` activates a proposal even when the slot is in use:

```rust
let new_sidechain_activated =
    (sidechain_slot_is_used && vote_count > used_sidechain_slot_activation_threshold && ...)
    || (!sidechain_slot_is_used && ...);
```

`AckSidechain::apply` then does `put_sidechain(slot, new)` (overwriting the incumbent) and deletes the proposal. `AckSidechain::undo` does `delete_sidechain(slot)`. Since the `AckSidechain` diff only stored `{id, activated}`, undo had no way to restore the overwritten sidechain, so disconnecting the activating block removes the slot entirely.

## Impact

- A chain reorg through such an activation block permanently deletes a sidechain that was active before the block; nodes with different reorg histories end up with divergent state (`connect_block` / `disconnect_block` are not inverse).
- If the deleted slot held a treasury, `ctip[slot]` persists while the slot becomes inactive. `handle_m5_m6` skips inactive slots, so the OP_DRIVECHAIN treasury output is no longer defended and can be spent — i.e. the reorg can orphan a treasury into an undefended state.

## Fix

Record the overwritten active sidechain in the `AckSidechain` diff (`replaced_active: Option<Sidechain>`, `serde(default)` for compatibility with previously-serialized diffs) and have `undo` restore it with `put_sidechain` instead of `delete_sidechain` when the slot was already in use.

## Test

`ack_sidechain_undo_restores_overwritten_active_sidechain`: active sidechain A in a slot, a competing proposal B activates over it, and after apply+undo the slot must again hold A (not be deleted).

### DIFF
--- a/lib/validator/dbs/diff.rs
+++ b/lib/validator/dbs/diff.rs
@@ -95,6 +95,13 @@ pub struct AckSidechain {
     /// true IFF the sidechain should be activated due to this ack causing the
     /// proposal to reach the vote threshold
     pub activated: bool,
+    /// When `activated` and the slot was *already* occupied by an active
+    /// sidechain, that prior sidechain (which `apply` overwrites). `undo` must
+    /// restore it instead of deleting the slot, otherwise a reorg of the
+    /// activating block permanently deletes the previously-active sidechain.
+    /// `serde(default)` keeps diffs stored before this field was added readable.
+    #[serde(default)]
+    pub replaced_active: Option<Sidechain>,
 }
 
 impl Diff for AckSidechain {
@@ -131,9 +138,25 @@ impl Diff for AckSidechain {
                 .active_sidechains
                 .sidechain()
                 .get(rwtxn, &self.id.sidechain_number)?;
-            let () = dbs
-                .active_sidechains
-                .delete_sidechain(rwtxn, &self.id.sidechain_number)?;
+            match &self.replaced_active {
+                // `apply` overwrote an already-active sidechain in this slot;
+                // restore it (leaving its pending_m6ids, which `apply` did not
+                // touch) instead of deleting the slot.
+                Some(prev) => {
+                    let () = dbs.active_sidechains.put_sidechain(
+                        rwtxn,
+                        &self.id.sidechain_number,
+                        prev,
+                    )?;
+                }
+                // The slot was empty before activation; `apply` created it, so
+                // undo removes it entirely.
+                None => {
+                    let () = dbs
+                        .active_sidechains
+                        .delete_sidechain(rwtxn, &self.id.sidechain_number)?;
+                }
+            }
             sidechain.status.activation_height = None;
             sidechain
         } else {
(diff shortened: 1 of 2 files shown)


torkelrogstad (approved):
Thank you!

lib/validator/dbs/diff.rs
```diff
@@ -95,6 +95,13 @@ pub struct AckSidechain {
     /// true IFF the sidechain should be activated due to this ack causing the
     /// proposal to reach the vote threshold
     pub activated: bool,
+    /// When `activated` and the slot was *already* occupied by an active
+    /// sidechain, that prior sidechain (which `apply` overwrites). `undo` must
+    /// restore it instead of deleting the slot, otherwise a reorg of the
+    /// activating block permanently deletes the previously-active sidechain.
+    /// `serde(default)` keeps diffs stored before this field was added readable.
+    #[serde(default)]
+    pub replaced_active: Option<Sidechain>,
```
Ash-L2L: This field is always `None`  if `activated` is `true` - it would be best to use an enum that makes it impossible to construct an invalid diff where `activated = false` and `replaced_active = Some(_)`.

Ash-L2L (changes requested):

1440000bytes:
When an M2 ack activates a new sidechain into a slot that's already in use, the incumbent gets overwritten, but its treasury (CTIP) and pending withdrawal bundles aren't cleared. They're stored per slot `ctip[slot]`, `treasury_utxo_count[slot]`, `slot_sequence_to_treasury_utxo[slot]`, and `pending_m6ids[slot]` (the latter only initialized when absent), so the new sidechain inherits the previous one's funds and bundles, and could spend the old treasury via an M6.

It needs the same capture-on-apply / restore-on-undo handling already added here, just extended to the treasury and `pending_m6ids` rather than only the sidechain entry. Will update this PR to cover the full slot state if that makes sense.

torkelrogstad (approved):

Ash-L2L:
>When an M2 ack activates a new sidechain into a slot that's already in use, the incumbent gets overwritten, but its treasury (CTIP) and pending withdrawal bundles aren't cleared. They're stored per slot ctip[slot], treasury_utxo_count[slot], slot_sequence_to_treasury_utxo[slot], and pending_m6ids[slot] (the latter only initialized when absent), so the new sidechain inherits the previous one's funds and bundles, and could spend the old treasury via an M6.

This is intended behavior. 

Approving this PR as-is

Ash-L2L (approved):

Ash-L2L (approved):

Ash-L2L (approved):

PR #394, opened 2026-06-16
wallet: apply unconfirmed deposit transaction post create

torkelrogstad:
Prior to this commit we would fail with RBF-looking errors due to the BDK wallet not having the correct view of our UTXOs.

### DIFF
--- a/lib/wallet/mod.rs
+++ b/lib/wallet/mod.rs
@@ -1443,6 +1443,36 @@ impl Wallet {
         }
         if broadcast_successfully {
             tracing::info!(%txid, "Broadcast deposit transaction successfully");
+
+            // Apply the unconfirmed deposit to the wallet so its funding input
+            // is marked spent. Otherwise a deposit created before this tx
+            // confirms can reselect the same UTXO, which Bitcoin Core rejects
+            // as an RBF replacement. Mirrors `send_wallet_transaction`.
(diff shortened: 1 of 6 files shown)


lib/wallet/mod.rs
```diff
@@ -1443,6 +1443,36 @@ impl Wallet {
...
+            let last_seen = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap();
+            let applied_changes = self
+                .inner
+                .write_wallet()
+                .await?
+                .with_mut(|wallet| {
+                    wallet.apply_unconfirmed_txs(vec![(tx, last_seen.as_secs())]);
+                    wallet.persist_async(&mut bdk_db_lock)
```
Ash-L2L, line 1461: Do we need to persist the wallet here?
We don't persist in [`accept_tx`](https://github.com/LayerTwo-Labs/bip300301_enforcer/blob/master/lib/wallet/cusf_block_producer.rs#L289), should we?

Ash-L2L (approved):
LGTM!

lib/wallet/mod.rs
torkelrogstad: The correct thing to do here seems to be persisting both places. Pushing a commit for that

PR #396, opened 2026-06-16
[DoS] Invalidate consensus-invalid blocks during sync instead of halting the enforcer

1440000bytes:
Found this bug by crafting a treasury-theft block and observing the enforcer halt while syncing it. The enforcer handles a consensus-invalid block (one the mainchain node accepted but that violates BIP300/301) **asymmetrically** depending on when it's seen:

- **Live path** (`cusf-enforcer-mempool`, block arrives via ZMQ): `connect_block` returns `ConnectBlockAction::Reject` and node is told to `invalidateblock` 
- **Sync path** (`handle_block_batch`, initial bootstrap / catch-up): the same `connect_block` error is propagated with a blind `?`, becomes a fatal `SyncError`, and aborts the whole task

So a block the live path cleanly rejects instead **bricks any enforcer that meets it during sync**:

- a fresh enforcer bootstrapping on a node whose active chain already contains such a block can never finish initial sync;
- an enforcer catching up after downtime / a deep reorg aborts on it.

This is cheap to trigger. A sidechain treasury is an `OP_DRIVECHAIN` output, which is **anyone-can-spend** at the script level and the node does not enforce BIP300 at consensus. So, anyone can mine a tx that spends the treasury to a plain output. The node accepts the block; `connect_block` correctly rejects it (`TreasurySpentWithoutNewCtip`), but the sync path turns that *non-fatal* rejection into a fatal halt. While the enforcer is down it can't invalidate the block, so the bad chain stands . The availability defect undermines the very protection the enforcer provides.

```
INFO  Synced block #108: `02e96351...` total_txs=1 sc_events=0
INFO  Synced block #109: `13d419da...` total_txs=2 sc_events=1
Error:   × exit-after-sync task failed
  ├─ Error handling transaction
  │   `a1d75302bbff246033bbf6bf2f86bad14142abcfcf1fbb55e475c3df64e1888c` in
  │   block `44cb0d671dd67b2b3ac50447f1597ed2d47cda9c5a23671db78849509db7def3`
  ├─ Error handling M5/M6
  ╰─ treasury UTXO for sidechain 0 is spent without creating a new treasury UTXO
```

## Fix

Make the sync path mirror the live path. `connect_block` errors already carry a fatal/non-fatal classification (`error_fatality`); the sync path now respects it:

- `handle_block_batch` surfaces a *non-fatal* `connect_block` rejection as a new recoverable `Sync::BlockRejected { block_hash, .. }` instead of a blind `?`
- `sync_blocks` calls `invalidateblock(block_hash)` on the node and returns `Ok` so the caller re-syncs to the reorged tip (exactly the live `ConnectBlockAction::Reject` behavior)
- Genuine fatal (infra/DB) errors still abort, unchanged.

## Tests

- New unit test `handle_block_batch_rejects_consensus_invalid_block_without_halting`:  asserts a consensus-invalid block is surfaced as the recoverable `Sync::BlockRejected` (carrying the offending hash), not a fatal error.
- Verified end-to-end against a drivechain-patched regtest node with a real treasury-theft block: pre-fix `--exit-after-sync` aborts (exit 1); with the fix the enforcer invalidates the block, reorgs it away, and exits cleanly.

### DIFF
--- a/lib/validator/task/error.rs
+++ b/lib/validator/task/error.rs
@@ -445,6 +445,15 @@ pub(in crate::validator) enum Sync {
     #[error("Block not in active chain: `{block_hash}`")]
     #[fatal(true)]
     BlockNotInActiveChain { block_hash: bitcoin::BlockHash },
+    /// A block that the mainchain node accepted, but that violates BIP300/301.
+    /// Not a fatal sync failure: the block is invalidated on the node, and the
+    /// sync is retried against the resulting tip.
+    #[error("Consensus-invalid block `{block_hash}` rejected during sync")]
+    #[fatal(false)]
+    BlockRejected {
+        block_hash: bitcoin::BlockHash,
+        source: Box<ConnectBlock>,
+    },
     #[error(transparent)]
     #[fatal(true)]
     CommitWriteTxn(#[from] rwtxn::error::Commit),
--- a/lib/validator/task/mod.rs
+++ b/lib/validator/task/mod.rs
@@ -2097,6 +2097,50 @@ mod tests {
         Ok(())
     }
 
+    // Regression test for the sync-path DoS: a block that the mainchain node
+    // accepts, but that is consensus-invalid per BIP300/301, must be reported as a
+    // recoverable `Sync::BlockRejected` carrying the offending block hash, so that
+    // the caller can invalidate it, rather than as a fatal error that halts a
+    // bootstrapping or catching-up enforcer.
+    #[test]
+    fn handle_block_batch_rejects_consensus_invalid_block_without_halting() -> Result<()> {
+        let (_dir, dbs) = create_test_dbs()?;
+        let mut rwtxn = dbs.write_txn().into_diagnostic()?;
+        let prev_hash = BlockHash::all_zeros();
+
+        let m8_tx = build_m8_tx(SidechainNumber(1), [0x42; 32], prev_hash);
+        let block = build_test_block(
+            prev_hash,
+            TestBlockParts {
+                extra_txs: vec![m8_tx],
+                ..Default::default()
+            },
+        );
+        let block_hash = block.block_hash();
+
+        dbs.block_hashes
+            .put_headers(&mut rwtxn, &[(block.header, 0)])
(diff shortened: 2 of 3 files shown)


lib/validator/task/error.rs
```diff
@@ -445,6 +445,16 @@ pub(in crate::validator) enum Sync {
...
     #[fatal(true)]
     BlockNotInActiveChain { block_hash: bitcoin::BlockHash },
+    // A block the mainchain node accepted but that violates BIP300/301. This is
+    // NOT a fatal sync failure: the sync loop invalidates the block on the node
+    // and re-syncs, mirroring the live `ConnectBlockAction::Reject` path.
+    #[error("Consensus-invalid block `{block_hash}` rejected during sync")]
+    #[fatal(false)]
+    BlockRejected {
+        block_hash: bitcoin::BlockHash,
+        #[source]
```
Ash-L2L: Nit: the `#[source]` attribute is not needed if the field is already named `source`

lib/validator/task/mod.rs
```diff
@@ -1750,8 +1763,37 @@ impl BlockHandler<'_> {
...
+                    main_rpc_client
+                        .invalidate_block(block_hash)
+                        .await
+                        .map_err(|err| error::Sync::JsonRpc {
+                            method: "invalidateblock".to_owned(),
+                            source: err,
+                        })?;
+                    // The node has reorged away from the rejected block; stop here
+                    // and let the caller re-sync to the new tip.
+                    return Ok(());
```
Ash-L2L: Returning `Ok(())` here is dangerous, the caller will have no indication that the sync failed. It would be better to propagate the error and handle it in `validator/cusf_enforcer.rs`.

Ash-L2L (changes requested):

1440000bytes:
I have addressed the review comments in the last push.

torkelrogstad:
Thank you for this submission and apologies for taking so long in responding to it. We fixed the issue with a slightly different approach in #549. Again, thanks for identifying the bug!

PR #400, opened 2026-06-18
validator: implement support for competing sidechain deposits

torkelrogstad:
A simpler version of #390

### DIFF
--- a/lib/validator/cusf_enforcer.rs
+++ b/lib/validator/cusf_enforcer.rs
@@ -22,11 +22,15 @@ use tokio_util::sync::CancellationToken;
 
 use crate::{
     errors::ErrorChain,
-    messages::parse_m8_tx,
+    messages::{parse_m8_tx, parse_op_drivechain},
     proto::mainchain::HeaderSyncProgress,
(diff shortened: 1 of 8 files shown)


lib/validator/cusf_enforcer.rs
```diff
@@ -259,6 +267,69 @@ fn connect_block_no_commit<'validator>(
...
+    tx: &Transaction,
+    txid: Txid,
+) -> Result<HashSet<Txid>, db::Error> {
+    let mut conflicts_with = HashSet::new();
+    for output in &tx.output {
+        let Ok((_, sidechain_number)) = parse_op_drivechain(output.script_pubkey.as_bytes()) else {
+            continue;
+        };
+        // An OP_DRIVECHAIN output for an inactive slot is an ordinary
+        // output, not a treasury UTXO
```
Ash-L2L, line 290: This is only true while the slot is inactive. The tx should still be tracked so that it can conflict with other txs if the slot becomes active

Ash-L2L (changes requested):

PR #402, opened 2026-06-19
Delete treasury utxo count key on disconnect of last treasury UTXO

1440000bytes:
`delete_ctip` (block disconnect) decrements `treasury_utxo_count` and writes the new value back but when the count reaches 0 it writes Some(0) instead of deleting the key. `put_ctip` creates the key at a count of 1 for the first treasury UTXO of a sidechain (M5 deposit, see bip300.md), so connecting then disconnecting the first deposit leaves a stale `treasury_utxo_count` of Some(0) where it should be absent. `disconnect_block` is not an exact inverse of `connect_block` for this key.

`put_ctip` reads the count with `unwrap_or(0),` so Some(0) and absent behave the same and sequence numbering stays correct. But it leaves dangling state after a reorg and breaks the connect/disconnect round-trip invariant.

The fix deletes the key when the decremented count reaches 0; the test asserts the count key is absent after disconnecting the only deposit (fails on master).


1440000bytes:
Duplicate of #387

PR #403, opened 2026-06-19
Reject undersized or truncated block file entries instead of panicking

1440000bytes:
`BlockFileParser::next_block `reads a 4‑byte size field straight from a blk*.dat and then does vec![0; size - 80]. A size < 80 (header length) underflows panic in debug, huge allocation/abort in release. The header decode (:245) and the block‑body read_exact (:249) also used .unwrap(), so a truncated/corrupt block file aborts the process rather than returning an error. When the enforcer is run with --main-blocks-dir, a single corrupt or half‑written block file therefore crashes it.

Fix: Add a `BlockSizeTooSmall` error, use checked_sub(80), and replace the two unwrap()s with the existing HeaderDecode / Io error variants. Test feeds a block with size = 0 and asserts next_block() returns an error instead of panicking (fails on master).

### DIFF
--- a/lib/validator/parse_block_files.rs
+++ b/lib/validator/parse_block_files.rs
@@ -58,6 +58,9 @@ pub enum ParseBlockFileError {
     #[error("Block size too large: {size} bytes")]
     BlockSizeTooLarge { size: u32 },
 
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
Thanks!

PR #416, opened 2026-06-27
[DoS] wallet: fix treasury value underflow when building withdrawal bundle suffix txs

1440000bytes:
`generate_suffix_txs` computed the post-withdrawal treasury value with unchecked `Amount` subtraction `(value - fee) - payout`, which panics and aborts the process when a bundle's payout plus fee exceeds the current treasury. A sidechain operator can reach this by broadcasting such a bundle via `BroadcastWithdrawalBundle` (it is not bounded against the treasury at submission time); once it is voted past the inclusion threshold, every block-template build panics, halting block production. The adjacent `BlindedM6::into_m6` already performs the same subtraction with `checked_sub`.

Remotely reachable DoS:

1. A sidechain operator calls `BroadcastWithdrawalBundle` (gRPC) with a bundle where `payout + fee > treasury value`. `BlindedM6::try_from` validates structure but cannot bound by the treasury (unknown at submit time) — confirmed no such check exists.
2. The enforcer's own miner auto-upvotes bundle index 0 every block until v`ote_count > withdrawal_bundle_inclusion_threshold`.
3. The next block-template build (`select_block_txs → generate_suffix_txs`) hits line 250 and panics → mining halts. The over-value tx never even needs to pass on-chain validation; the crash is in the producer.

This routes the suffix-tx computation through a checked helper that returns `AmountUnderflowError` instead, with a regression test.

### DIFF
--- a/integration_tests/integration_test.rs
+++ b/integration_tests/integration_test.rs
@@ -6,20 +6,22 @@ use bip300301_enforcer_lib::{
         self,
         common::{ConsensusHex, Hex},
         mainchain::{
-            CreateDepositTransactionRequest, CreateDepositTransactionResponse,
-            CreateNewAddressRequest, CreateSidechainProposalRequest, GetInfoRequest,
+            BroadcastWithdrawalBundleRequest, CreateDepositTransactionRequest,
+            CreateDepositTransactionResponse, CreateNewAddressRequest,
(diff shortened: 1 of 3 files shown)


torkelrogstad (approved):
Thanks! I rebased this, and expanded the integration tests.

PR #417, opened 2026-06-27
[DoS] server: fix sidechain proposal age underflow in get_sidechain_proposals

1440000bytes:
`get_sidechain_proposals` computes each proposal's `proposal_age` with an unchecked `u32` subtraction:

```rust
proposal_age: wrap_u32(mainchain_tip_height - sidechain.status.proposal_height),
```

When a sidechain proposal retained from a previous, non-active sync has a `proposal_height` greater than the current `mainchain_tip_height`, this subtraction underflows: it panics under overflow checks, or wraps to a garbage age in release. `get_sidechains` returns every stored proposal, so a stale proposal above the active tip is enough to trigger it.

`get_sidechain_proposals` takes an empty request, so any caller can reach this path and crash the enforcer.

## Fix

The same computation in `validator::task` already guards against this with `saturating_sub` (a proposal can have a height above the active tip after a reorg/resync). This routes the gRPC handler through a shared `proposal_age` helper that saturates instead of underflowing and adds a regression test.

### DIFF
--- a/lib/server/validator/grpc.rs
+++ b/lib/server/validator/grpc.rs
@@ -31,6 +31,14 @@ use crate::{
     types::Thresholds,
 };
 
+/// Age of a sidechain proposal at the given mainchain tip height. A proposal
+/// retained from a previous sync can have a `proposal_height` above the active
+/// tip, so this saturates instead of underflowing, matching the same
+/// computation in `validator::task`.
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
Similar to #392 I don't really see how this can be triggered in prod. But the change is safe and sensical. Thanks!

PR #418, opened 2026-06-28
[DoS] validator: reject blocks with no coinbase instead of panicking in connect_block

1440000bytes:
`connect_block` reads the coinbase with an unchecked index:

```rust
let coinbase = &block.txdata[0];
```

A block whose `txdata` is empty makes this panic with `index out of bounds: the len is 0 but the index is 0`, aborting the enforcer. There is no emptiness check and no merkle validation before the index.

The block-file sync path (`--main-blocks-dir`) is the trigger. `BlockFileParser::next_block` reads a `u32` size, an 80-byte header, then `txdata_len = size.checked_sub(80)` (the guard added in #403). With `size = 81` and a single `0x00` body byte, `txdata_len = 1` and `Vec::<Transaction>::consensus_decode(&[0x00])` returns `Ok(vec![])` an empty transaction list. The block hash is computed from the header alone, so a crafted `blk*.dat` can copy a genuine, expected block's real 80-byte header and pass through to `connect_block`, which then panics.

This is distinct from #403 (which only guards `size < 80`) and #396 (which reclassifies returned errors and cannot catch a panic).

## Fix

Reject a block with no coinbase via a new non-fatal `ConnectBlock::NoCoinbase` error instead of indexing. Includes a regression test that connects an empty-`txdata` block and asserts it is rejected rather than panicking.

### DIFF
--- a/lib/validator/task/error.rs
+++ b/lib/validator/task/error.rs
@@ -375,6 +375,9 @@ pub(in crate::validator) enum ConnectBlock {
     #[error("Multiple blocks BMM'd in sidechain slot {}", .sidechain_number.0)]
     #[fatal(false)]
     MultipleBmmBlocks { sidechain_number: SidechainNumber },
+    #[error("Block has no transactions (missing coinbase)")]
+    #[fatal(false)]
+    NoCoinbase,
     #[error(transparent)]
     #[fatal(true)]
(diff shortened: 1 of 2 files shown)


torkelrogstad (approved):
Thanks!

PR #419, opened 2026-06-28
Reject M5 deposit with no address OP_RETURN output

1440000bytes:
BIP300 requires the address `OP_RETURN` output immediately after the treasury UTXO in an M5 deposit, and states:

> If an M5 transaction has a UTXO treasury output and no address OP_RETURN output, then it MUST be considered invalid by the BIP300 enforcer software.

https://github.com/LayerTwo-Labs/bip300_bip301_specifications/blob/master/bip300.md#L581-L582

`handle_m5_m6` instead derives the address defensively and never rejects:

```rust
let address = if let Some(address_output) =
    transaction.output.get(new_ctip.outpoint.vout as usize + 1)
{
    crate::messages::try_parse_op_return_address(&address_output.script_pubkey)
        .unwrap_or_default()
} else {
    Vec::new()
};
```

When the output after the treasury is absent, or is not a valid `OP_RETURN <push>` (e.g. a P2WPKH payment, or an `OP_RETURN` with multiple pushes), `try_parse_op_return_address` returns `None`, `.unwrap_or_default()` yields an empty address, and the deposit is still accepted and the CTIP advanced. A miner can mine such a transaction; a spec-compliant enforcer rejects it, so the two diverge on block validity.

## Fix

Reject the deposit with a new non-fatal `HandleM5M6::MissingDepositAddress` error when the address output is absent or does not parse as an `OP_RETURN` push (an `OP_RETURN` with an empty push is still accepted, since address contents are up to the sidechain). Includes a regression test; the existing valid-deposit tests still pass.

### DIFF
--- a/integration_tests/test_invalid_block.rs
+++ b/integration_tests/test_invalid_block.rs
@@ -1,13 +1,13 @@
-use std::time::Duration;
+use std::{str::FromStr as _, time::Duration};
 
 use bip300301_enforcer_lib::{
     bins::CommandExt as _,
     messages::{M1ProposeSidechain, M2AckSidechain, M4AckBundles, M7BmmAccept},
-    types::{BmmCommitment, SidechainDescription},
+    types::{BmmCommitment, SidechainDescription, op_drivechain_script},
 };
 use bitcoin::{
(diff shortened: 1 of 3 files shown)


torkelrogstad (approved):
Rebased this and added integration test coverage. Thanks!

PR #421, opened 2026-06-28
Reject non-minimal pushes when parsing coinbase messages

1440000bytes:
`CoinbaseMessage::parse` decoded the OP_RETURN payload with `Script::instructions()`, which does **not** enforce minimal pushes. As a result a coinbase BIP300/301 message (M1–M4, M7) encoded with a non-minimal push opcode e.g. `OP_PUSHDATA1 <len> <data>` for a payload that fits in a direct `OP_PUSHBYTES_n` parsed identically to its canonical form:

```rust
let mut instructions = script.instructions();
...
let Some(Ok(Instruction::PushBytes(data))) = instructions.next() else { ... };
```

Accepting a second, non-canonical encoding of the same message is a malleability / divergence risk.

## Fix

Decode with `Script::instructions_minimal()`, which rejects non-minimal pushes, so each coinbase message has a single canonical encoding. The canonically-encoded messages the wallet produces (`ScriptBuf::try_from(..)` uses minimal pushes) are unaffected; all existing parse/round-trip tests pass.

A regression test encodes a valid M2 message, confirms it parses, then re-wraps the same payload with a non-minimal `OP_PUSHDATA1` and asserts `CoinbaseMessage::parse` rejects it.

### DIFF
--- a/lib/messages.rs
+++ b/lib/messages.rs
@@ -8,7 +8,7 @@ use bitcoin::{
         OP_TRUE,
         all::{OP_PUSHBYTES_1, OP_RETURN},
     },
-    script::{Instruction, Instructions, PushBytesBuf},
+    script::{Instruction, PushBytesBuf},
 };
 use byteorder::{ByteOrder, LittleEndian};
 use miette::Diagnostic;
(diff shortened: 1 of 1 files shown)


torkelrogstad:
I don't believe this exists as a requirement in the spec. If we were to have such a limitation, it'd also have to go in there. And not sure that is something we want? But open to your thoughts @1440000bytes

1440000bytes:
> I don't believe this exists as a requirement in the spec. If we were to have such a limitation, it'd also have to go in there. And not sure that is something we want? But open to your thoughts @1440000bytes

On second thought, I think you're right. It's only a consensus problem if another implementation parses it differently.

torkelrogstad:
Closing this, then

PR #422, opened 2026-06-28
Reject deposit value that overflows the treasury total

1440000bytes:
`create_m5_deposit_output` computed the new treasury value with an unchecked add:

```rust
value: old_ctip_amount + deposit_amount,
```

`bitcoin::Amount`'s `+` is `checked_add(..).expect(..)`, so this panics when the sum overflows. `deposit_amount` comes from the gRPC `create_deposit_transaction` `value_sats` field, which is only checked for non-zero `Amount::from_sat` does not bound it, so a caller passing a large `value_sats` (such that `old_ctip + value` exceeds the max) crashes the request.

## Fix

Use a checked add and return `AmountOverflowError`, propagated through `create_deposit_op_drivechain_output` - `create_deposit` (new `CreateDeposit::OutputAmountOverflow` variant), so an out-of-range deposit is rejected with an error instead of panicking. Adds a regression test.

### DIFF
--- a/lib/messages.rs
+++ b/lib/messages.rs
@@ -1245,6 +1245,18 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn create_m5_deposit_output_rejects_overflow() {
+        // A deposit value that overflows the treasury total must error rather
+        // than panic on the addition.
+        let result = create_m5_deposit_output(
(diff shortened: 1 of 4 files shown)


torkelrogstad (approved):
Thank you!

PR #427, opened 2026-06-30
wallet: integrate reusable-payments crate

mplatt8:
95%+ of this diff is regenerated protobuf (lib/proto/generated/)


torkelrogstad:
This should give a clearer error message, ideally matching what we already do for insufficient funds on normal sends

```
❯ buf curl --emit-defaults --http2-prior-knowledge \
        -d '{"paymentCode":"PM8TJLjTLM4VSZdvvjWH2KkigLxRGA3cFqzAGtppbQTks5E6U7YrhGvMHcFwnvc2nS8xwMZtjGzrZh9edBQfYqCBexJeWnjkPmxPTzYnRwwMZxjRwf1R","amountSats":"250000","feeSatPerVbyte":"2"}' \
        http://localhost:50051/cusf.mainchain.v1.WalletService/SendToBip47PaymentCode
{
   "code": "internal",
   "message": "Bitcoin Core RPC error during reusable-payments scan: notification psbt: Insufficient funds: 0 BTC available of 0.00000819 BTC needed"
}
```

lib/server/wallet/grpc.rs
```diff
@@ -791,4 +822,281 @@ impl WalletService for crate::wallet::Wallet {
...
+        ))
+    }
+
+    async fn rescan_reusable_payments(
+        &self,
+        _ctx: RequestContext,
+        request: ServiceRequest<'_, crate::proto::mainchain::RescanReusablePaymentsRequest>,
+    ) -> ServiceResult<crate::proto::mainchain::RescanReusablePaymentsResponse> {
+        let req = request.to_owned_message();
+        self.rescan_reusable_payments(req.from_height)
```
torkelrogstad: Couple of issues here: 

1. What happens if the caller provides a height that is below the wallet birthday? Do we return an error?
2. What happens if the the caller provides 0? Would seem sensible to scan from the wallet birthday by default
3. Should we validate that the height is below the chain tip?

torkelrogstad: The integration tests added a demonstration of a potential issue with a below birthday rescan attempt, btw

integration_tests/setup.rs
```diff
@@ -448,6 +448,9 @@ pub struct SetupOpts<
     pub bitcoind_args: BitcoindArgs,
     pub bitcoind_kind: BitcoindKind,
     pub enforcer_args: EnforcerArgs,
+    /// Create the enforcer wallet from this seed file instead of a random
+    /// mnemonic, so tests can derive the wallet's keys independently.
+    pub wallet_seed_file: Option<PathBuf>,
```
Ash-L2L, line 453: Can we specify a seed value instead of a file or mnemonic phrase?

lib/server/wallet/grpc.rs
```diff
@@ -171,6 +171,38 @@ where
     Ok(sidechain_proposal)
 }
 
+fn bip47_version_from_proto(
+    raw: buffa::EnumValue<crate::proto::mainchain::Bip47Version>,
+) -> Result<crate::wallet::reusable_payments::bip47::Version, ConnectError> {
+    use crate::{proto::mainchain::Bip47Version as P, wallet::reusable_payments::bip47::Version};
+    match raw.as_known() {
+        Some(v) if v == P::V1 => Ok(Version::V1),
```
Ash-L2L, line 179: This match looks odd, you should be  able to do

```rust
match raw.as_known() {
        Some(P::V1) => Ok(Version::V1),
        Some(P::V3) => Ok(Version::V3),
        _ => Err(ConnectError::invalid_argument(
            "Bip47Version must be V1 or V3",
        )),
```

Ash-L2L (changes requested):

PR #433, opened 2026-07-04
Avoid underflow in block index data position for corrupt entries

1440000bytes:
`CDiskBlockIndex::adjusted_data_pos` computes a block's data position with an unchecked subtraction:

```rust
pub fn adjusted_data_pos(&self) -> Option<u64> {
    const ADJUST_BY: u64 = 8;
    self.data_pos.map(|pos| pos - ADJUST_BY)   // underflows if data_pos < 8
}
```

`data_pos` is read (as a Core VarInt) from a `blocks/index` LevelDB value, and this runs on the `--main-blocks-dir` fast-sync path (`fetch_block_index` → `adjusted_data_pos()` → `set_offset`). A corrupt or crafted index entry with `data_pos < 8` underflows: under overflow-checks (the test suite / CI run in debug) it **panics**; in release it **wraps** to a near-`u64::MAX` offset and returns `Some(bogus)`, which defeats the caller's `.ok_or(MissingDataPos)?` guard → a bad `set_offset` and a failed/skipped block during sync.

Same class as the block-file size underflow fixed in #403.

## Fuzzing logs

A libFuzzer harness over `CDiskBlockIndex::deserialize` + `adjusted_data_pos`, found in ~14k executions:

```
thread panicked at parse_block_files.rs: attempt to subtract with overflow
==12462== ERROR: libFuzzer: deadly signal
```

### DIFF
--- a/lib/validator/parse_block_files.rs
+++ b/lib/validator/parse_block_files.rs
@@ -607,7 +607,10 @@ impl CDiskBlockIndex {
     pub fn adjusted_data_pos(&self) -> Option<u64> {
         // Magic bytes + size? Found by trial and error
         const ADJUST_BY: u64 = 8;
-        self.data_pos.map(|pos| pos - ADJUST_BY)
+        // `checked_sub` so a corrupt index entry with `data_pos < ADJUST_BY`
+        // yields `None` (handled as a missing data position) instead of
+        // underflowing.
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
Thank you! Rebased the PR.

PR #34, opened 2024-10-11
wallet gRPC scaffold - this time against the correct branch

torkelrogstad:
I tried addressing your comments @Ash-L2L .

PR #459, opened 2026-07-16
BIP 360 CUSF

cryptoquick:
### Core enforcement (enforcer / `bip360` feature)
- [x] BIP 360 / P2MR validation on stock bitcoind (CUSF: no Core patches for the claim)
- [x] Feature flag isolation (`bip360`; default path stays drivechain-clean)
- [x] `quantum` → `pqc` rename + module layout under `lib/validator/pqc/`
- [x] Overload-model leaves (`OP_CHECKSIG` size-gated; `OP_SUBSTR` forbidden in enforcer)
- [x] Cross-block P2MR prevout DB (`p2mr_utxos`)
- [x] Schnorr, ML-DSA-44, SLH-DSA-SHA2-128s verify paths
- [x] Hybrid EC+SLH and kitchen-sink (triple-algo) leaves
- [x] Multi-leaf algorithm-exclusion P2MR
- [x] DoS limits (witness stack, per-input WU, PQC verify budget)
- [x] `invalidateblock` path for invalid BIP 360 blocks (Scenario B / demo-b)

### Tests & demos
- [x] PQC unit suite (`pqc::` — 123+ tests)
- [x] Overload construction vectors (`p2mr_overload_construction.json`)
- [x] Block matrix live: **33/33 PASS** (`just bip360-block-matrix`)
- [x] Dual-node P2P E2E trial #34 PASS (`just bip360-p2p-e2e`; electrs unblocked)
- [x] Kitchen-sink block trials (valid + tamper)
- [x] **Tier A kitchen-sink demo** PASS: stock Alice + cryptoquick P2MR Bob
      (`just setup-p2mr` + `just bip360-kitchen-sink-tier-a`; jbride/bitcoin#2 head)
- [x] Mempool honesty: funding OK; v2 *spends* use submitblock on stock (and Bob still dialect-mismatches enforcer format → fallback)
- [x] `p2mr_signer` example + roundtrip tests
- [x] `just demo-a` / `just demo-b` recipes

### Infra & deps
- [x] electrs v3.2.0 build (`CXXFLAGS=-include cstdint`) for wallet+mempool path
- [x] Merge LayerTwo-Labs `master` into `bip360-cusf`
- [x] ZMQ / CUSF / mempool findings doc (`docs/ZMQ_CUSF_BIP360_FINDINGS.md`)
- [x] Mempool relay policy + CUSF-BIP360 docs updated
- [x] Tier A alignment note (`docs/TIER_A_P2MR_ALIGNMENT.md`; dual-valid dialects)
- [x] P2MR peer source of truth: **cryptoquick:p2mr** (jbride#2 head), ZMQ-enabled build
- [x] `bitcoinpqc` available as published **0.4.0** (enforcer still on git pin for now)

### Packaging prep (not necessarily merged upstream yet)
- [x] Upstream PR package drafted (`docs/UPSTREAM_PR.md`)
- [x] Status / design / artifacts / criteria audit at workspace root
- [x] Signet workshop docs prepared (live workshop still human)

---

## Remaining (P2MR packaging)

- [x] **Drop Kellnr for the rust-bitcoin P2MR fork**
      - Today: `bitcoin-p2mr-pqc` `0.32.6-p2mr-pqc.1` via `crates.denver.space` + `.cargo/config.toml`
      - Target: git-pin (e.g. `cryptoquick/rust-bitcoin`) and/or a public publish story so the L2 Labs PR does not depend on Kellnr
      - Optional follow-on: re-pin `bitcoinpqc` from git → crates.io `0.4.0` now that it’s published

---

## Explicitly deferred / HITL

- [ ] SHRINCs / WOTS+ (deferred)
- [x] Enforcer ↔ cryptoquick Core leaf/control-block interop (mempool path still `Witness program hash mismatch`)


cryptoquick:
We used everything currently available with libbitcoinpqc. Would love to add WOTS+C, SHRINCs, XMSS, and SQIsign 2.0, and maybe also an algorithm of my own one day, but I think this is good enough work for now.

psztorc:
This is very exciting.

I think we should build a second enforcer:

- Bitcoin Core (untouched)
- Bip300_301 enforcer
- Bip360_345_420 enforcer 
- BitWindow (etc)

To tie it all together

cryptoquick:
Dug into this feedback a little and the bottleneck will be ZMQ / RPC coordination.

We could do this feature binary arrangement, but they will need to talk to a central cusf coordinator node also, and there will need to be local IPC consensus.

I have designed this architecture already and will work to build and test it. It's not a major refactor and shouldn't take too long.

cryptoquick:
So, the problem with the approach you described is that it doesn't scale well with how cargo compiles binary feature flags.

Additionally, you can't just run multiple daemons with multiple ZMQ/RPC connections to bitcoind for a couple reasons.

I've rearchitected it to work more like this:

ae57352575e92bf9a76019e263823da780734f3e7031d7a7a2f5984f4b3cb7d6-5-65535-o-r3


8bfd9c6d47aa38ef40e080d961e521de0f571ec5d99a860f886659f980af7abd-5-65535-o-r3

Let me know if you have any further thoughts or questions.

torkelrogstad:
There's currently lots of non BIP300/301-related functionality in the enforcer for general chain management. That is a natural thing to reuse for a BIP360 enforcer. However not sure if introducing BIP360 functionality itself is a good fit for this repo. `cusf-enforcer-mempool` has functionality for composing together multiple different enforcers, where each enforcer has to accept a transaction for it to be valid. One idea would be to run `cusf-enforcer-mempool` as a standalone process, instead of in-process as we currently do in this repo. That way multiple enforcers could talk to the same running `cusf-enforcer-mempool`. Another idea would be to see how much general chain management functionality we could extract from `bip300301_enforcer`, and use that as the basis for a BIP-agnostic enforcer that serves as the basis for BIP300/301, BIP360, BIP X enforcement. These ideas are not mutually exclusive!

One issue with splitting of general parts from `bip300301_enforcer` is that it is currently under rather heavy development, which will continue for the foreseeable future. Splitting of parts would add to the burden of maintenance and development. So perhaps that'd be a thing better done in a few months?

With that being said, really cool that you're working on this! That's exciting.

cryptoquick:
Um, okay. Well, I did realize we needed to split out daemon comms (ZMQ/RPC) into its own daemon, so maybe that can be combined with the mempool stuff you mentioned. I can help with this refactoring work also.

Nuhvi:
Excuse me for asking this basic question here, because I don't know where else to ask, and the CUSF paper doesn't answer this question:
How are miners supposed to know it is time to enforce this or any other CUSF if there is no signaling to coordinate? are they supposed to reject blocks and see if they get orphaned? Is there any signaling mechanism that is taken for granted and I am not aware of?

PR #461, opened 2026-07-21
Fix successor CTIP output index for chained M6 withdrawals

ekulkisnek:
## Summary

Fix successor CTIP tracking after constructing an M6 withdrawal transaction.

BIP300 places the replacement treasury output at `vout 0`. The block producer previously selected the transaction’s last output, which may be a withdrawal payout.

## Impact

When an M6 had multiple outputs, the block producer could use a payout output as the next CTIP. A subsequent M6 constructed in the same block would then spend the wrong output and produce an invalid transaction chain.

The block producer now always records:

```text
successor CTIP = M6 transaction ID:vout 0
```

## Testing

Added a regression test using a multi-output M6 transaction that verifies:

- the successor transaction ID matches the M6;
- the successor output index is zero;
- the updated treasury value is preserved.

Focused test: `cargo test -p bip300301_enforcer_lib successor_ctip_is_always_m6_output_zero --lib`

### DIFF
--- a/lib/block_producer/coinbase.rs
+++ b/lib/block_producer/coinbase.rs
@@ -260,6 +260,18 @@ impl BlockProducer {
             .ok_or(AmountUnderflowError)
     }
 
+    /// Return the CTIP created by a successful M6. BIP300 fixes the treasury
+    /// output at index zero; payout outputs always follow it.
+    fn successor_ctip(m6: &Transaction, value: Amount) -> Ctip {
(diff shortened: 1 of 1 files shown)


torkelrogstad:
Thanks for the PR. I incorporated this into https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/543, where you're credited in the commit authorship.

PR #479, opened 2026-07-26
Broadcast sync events only after committing the batch

1440000bytes:
`handle_block_batch` broadcasts each block's `ConnectBlock` or `DisconnectBlock` event from inside the batch loop, while the caller commits the shared `rwtxn` only after the whole batch has been handled. The comment immediately above the `try_broadcast` call states the invariant it breaks:

> Events should only ever be sent after committing DB txs, see https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/185

If a later block in the batch fails, the error propagates and the transaction is aborted, but the earlier blocks' events have already been delivered. Subscribers see a `ConnectBlock` for state that was rolled back, and see it again, out of order, when the batch is retried. The reorg path in `sync_blocks` has the same shape: it disconnects blocks in a loop that broadcasts as it goes, then commits afterwards. So does the real time `disconnect_block` in `cusf_enforcer.rs`.

This is a notification ordering defect only. The validator DB rolls back correctly, so there is no consensus, funds or mining impact.

## Fix

`handle_block_batch` and `disconnect_block` now collect their events into a `Vec` instead of broadcasting them. A new `handle_block_batch_and_commit` takes the `RwTxn` by value, commits it, and only then broadcasts, so the batch call sites cannot get the order wrong. The sync reorg loop and the real time disconnect path do the same: commit, then broadcast.

### DIFF
--- a/lib/validator/cusf_enforcer.rs
+++ b/lib/validator/cusf_enforcer.rs
@@ -423,8 +423,10 @@ impl CusfEnforcer for Validator {
     ) -> Result<DisconnectBlockAction, Self::DisconnectBlockError> {
         let mut rwtxn = self.dbs.write_txn()?;
         let handler = BlockHandler::new(&self.dbs, self.network, self.network_params);
-        let () = handler.disconnect_block(&mut rwtxn, &self.events_tx, block_hash)?;
+        let mut events = Vec::new();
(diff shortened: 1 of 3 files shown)


torkelrogstad:
Thank you. I opened up a rebased version of this, with your authorship retained.

PR #480, opened 2026-07-26
wallet: do not fail block generation when BMM request cleanup fails

1440000bytes:
`generate_block` mines and submits the block, and only then deletes the BMM requests it consumed:

```rust
let block_hash = self.mine(&coinbase_outputs, transactions).await?;
self.inner.producer.db()
    .delete_bmm_requests(&mainchain_tip, &block_hash)
    .await
    .map_err(error::GenerateBlock::DeleteBmmRequests)?;
```

`mine` has already called `submitblock`, which is irreversible. If the delete then fails, the `?` reports the whole block generation as failed even though a block was produced, and the `bmm_requests` rows survive keyed to the old tip. A caller that reacts to that error by retrying reads the same rows back through `get_bmm_requests(old_tip)` and can mine a second mainchain block carrying the same M7 BMM accept, which breaks the at most one acceptance per sidechain block expectation. The window is bounded by how long the validator takes to advance its tip.

Reaching the double commit needs a genuine SQLite failure plus a retry inside that window, so it is not easy to trigger. Reporting a successful, irreversible operation as failed is wrong on its own.

## Fix

Move the cleanup into `Wallet::consume_bmm_requests`, which logs a delete failure instead of returning it. The block is already on the network, so generation reports success, and no caller is invited to retry an operation that already happened. The rows stay keyed to the old tip, which is the same state that #439's restore path already treats as correct after a disconnect.

The `GenerateBlock::DeleteBmmRequests` error variant is now unused and removed.

### DIFF
--- a/lib/wallet/error.rs
+++ b/lib/wallet/error.rs
@@ -1072,8 +1072,6 @@ impl ToStatus for GenerateSignetBlock {
 pub enum GenerateBlock {
     #[error(transparent)]
     CoinbaseBuilder(#[from] CoinbaseMessagesError),
-    #[error("failed to delete BMM requests")]
-    DeleteBmmRequests(#[source] rusqlite::Error),
     #[error(transparent)]
     GenerateCoinbaseTxouts(#[from] GenerateCoinbaseTxouts),
     #[error(transparent)]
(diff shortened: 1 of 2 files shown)


torkelrogstad:
Thank you. Closing in favor of #541. Your git authorship is retained

PR #481, opened 2026-07-26
Drop stale sidechain ACKs before adding auto-ACKs

1440000bytes:
In `extend_coinbase_txouts` the `ack_all_proposals` auto-ACK pass runs before the pass that validates and deletes stored ACKs, and it decides whether an auto-ACK is needed by looking at `sidechain_number` alone:

```rust
if !sidechain_acks.iter().any(|ack| ack.sidechain_number == sidechain_number) {
    sidechain_acks.push(SidechainAck { sidechain_number, description_hash: ... });
}
```

So when the wallet still holds a stored ACK for slot N with a stale description hash, and the only pending proposal for slot N is a replacement with a different hash, the stale row makes the auto-ACK pass believe the slot is already covered. No ACK is emitted for the replacement. The stale row is only removed by the validation pass further down, so the replacement's M2 appears in the second generated template at the earliest.

If the replacement proposal is near the edge of its voting window, one missed ACK can be the difference between activation and expiry. Producer local, no reorg needed.

## Fix

Split the decision into `select_sidechain_acks`, which first partitions the stored ACKs into the ones that still match a pending proposal and the stale ones, and only then adds an auto-ACK for any pending proposal left uncovered. A stale row can no longer stand in for a replacement, and the deletion of stale rows happens after that decision rather than before.

### DIFF
--- a/lib/block_producer/coinbase.rs
+++ b/lib/block_producer/coinbase.rs
@@ -392,9 +392,24 @@ impl BlockProducer {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashMap;
+
     use bitcoin::Amount;
 
-    use crate::{block_producer::BlockProducer, types::AmountUnderflowError};
+    use crate::{
+        block_producer::BlockProducer,
+        types::{
+            AmountUnderflowError, SidechainAck, SidechainDescription, SidechainNumber,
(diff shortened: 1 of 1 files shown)


torkelrogstad (approved):
Thank you. Rebased.

PR #483, opened 2026-07-27
Support nodes running under `assumeutxo`, clearly reject pruned nodes

torkelrogstad:
1. Pruned nodes are rejected with a clear error message, instead of cryptically failing when block data is not available.
2. Support nodes running with `assumeutxo` by waiting for blocks to be processed in the background before syncing.


lib/validator/task/chainstates.rs
```diff
@@ -0,0 +1,122 @@
...
+        let background_height = self
+            .chainstates
+            .iter()
+            .filter(|chainstate| chainstate.snapshot_blockhash.is_none())
+            .filter_map(|chainstate| chainstate.blocks)
+            .max()
+            // A background chainstate that hasn't connected anything yet
+            // (or reports an empty object) has served nothing but genesis.
+            .unwrap_or(0);
+        Some(background_height)
```
1440000bytes, line 49: This returns the background chainstate height as one flat ceiling for all block data, but a node that is background validating a snapshot serves everything from the snapshot base upwards for the entire background sync. The snapshot chainstate downloads and stores those blocks on its way to the tip, so `getblock` answers for them immediately.

Checked on regtest against Bitcoin Core master (v31.99.0-67efced1fc83), snapshot base 299, snapshot chainstate synced to 399, background chainstate still at genesis.

```
getchainstates
  chainstates[0]   blocks 0     validated true
  chainstates[1]   blocks 399   snapshot_blockhash 0c552ced...   validated false

getblock height   1    Block not available (not fully downloaded)
getblock height 149    Block not available (not fully downloaded)
getblock height 299    Block not available (not fully downloaded)
getblock height 300    available
getblock height 349    available
getblock height 399    available
```

Feeding that exact response to `block_data_available_height` gives `Some(0)`, so `sync_blocks` refuses heights 300 to 399 and polls `getchainstates` once a second until background validation finishes, for blocks the node hands over right now.

A fresh enforcer database does not hit this, since it needs the low blocks first anyway. It bites when the enforcer tip is already at or above the snapshot base, which is what you get by re-bootstrapping an existing node with `loadtxoutset` and keeping the enforcer data dir. The enforcer then has gRPC up and looks healthy while connecting no blocks, so BIP300 ACK counting and BIP301 BMM validation quietly stop.

Suggested fix, model the gap rather than a ceiling. `getchainstates` gives the snapshot blockhash, and `sync_blocks` already runs after `sync_headers`, so the base height can come from the enforcer's own header db without another RPC.

```rust
/// The snapshot base blockhash and the background chainstate height, while
/// an assumeutxo snapshot is still being validated in the background.
/// `None` if every block on the active chain is available.
pub(super) fn background_sync(&self) -> Option<(bitcoin::BlockHash, u32)> {
    let snapshot_blockhash = self.chainstates.iter().find_map(|chainstate| {
        chainstate
            .snapshot_blockhash
            .filter(|_| chainstate.validated != Some(true))
    })?;
    let background_height = self
        .chainstates
        .iter()
        .filter(|chainstate| chainstate.snapshot_blockhash.is_none())
        .filter_map(|chainstate| chainstate.blocks)
        .max()
        .unwrap_or(0);
    Some((snapshot_blockhash, background_height))
}
```

Then wait only for heights inside the gap, instead of everything above the background height.

```rust
let snapshot_base_height = dbs.block_hashes.height().get(&rotxn, &snapshot_blockhash)?;
let must_wait = background_height < next_height && next_height < snapshot_base_height;
```

Worth noting the `unwrap_or(0)` on its own too. Any shape it does not expect turns into "only genesis is available", which `sync_blocks` reads as an indefinite wait rather than an error.

The integration test does not cover this because the fixture has `base_height` 299 on a 299 block chain, so the snapshot base is the chain tip and there is never a block above it. Extending the fixture chain past the base and starting the enforcer with a db already at the base reproduces the stall.

1440000bytes:
The two failing unit tests: [chainstates.rs](https://github.com/1440000bytes/bip300301_enforcer/blob/assumeutxo-above-snapshot-base-repro/lib/validator/task/chainstates.rs#L124-L200)
The end-to-end repro: [test_node_requirements.rs](https://github.com/1440000bytes/bip300301_enforcer/blob/assumeutxo-above-snapshot-base-repro/integration_tests/test_node_requirements.rs#L524-L727)

lib/validator/task/chainstates.rs
torkelrogstad: Thank you! Incorporated.

PR #529, opened 2026-08-12
wallet: reject withdrawal bundles without a CTIP

### LINKED ISSUE #97: Reject Blinded M6 requests if no CTIP exists

ekulkisnek:
## Problem

An active sidechain may not have a treasury UTXO yet. `BroadcastWithdrawalBundle` currently accepts and persists a blinded M6 in that state even though there is no CTIP from which a withdrawal transaction can be constructed.

## Fix

Check the validator CTIP immediately after the existing active-sidechain gate. If none exists, return `FailedPrecondition` with a specific missing-treasury message.

## Tests

- Add an integration regression that activates a sidechain without making a deposit and verifies the RPC is rejected.
- Update the existing positive blinded-M6 and sidechain-ACK-policy fixtures to establish a confirmed CTIP before proposing a bundle; retain their storage, M3, and M4 assertions.
- `cargo +nightly fmt --all -- --check`
- `cargo check --all-targets`
- `cargo clippy --all-targets -- --deny warnings`
- strict nightly `unqualified_local_imports` clippy pass
- `cargo test --tests` (136 tests)
- focused `blinded_m6_requires_ctip` integration test
- focused `blinded_m6_roundtrip` integration test
- focused `sidechain_ack_policy` integration test

## Scope

This changes only wallet RPC ingestion for active sidechains without a CTIP. Inactive-sidechain rejection and funded-sidechain paths remain covered.

Closes #97

### DIFF
--- a/integration_tests/test_blinded_m6_roundtrip.rs
+++ b/integration_tests/test_blinded_m6_roundtrip.rs
@@ -13,7 +13,7 @@ use either::Either;
 use futures::channel::mpsc;
 
 use crate::{
-    integration_test::{activate_sidechain, propose_sidechain},
+    integration_test::{activate_sidechain, deposit, fund_enforcer, propose_sidechain},
     mine::mine_generateblocks_check,
     setup::{DummySidechain, PostSetup, Sidechain as _},
 };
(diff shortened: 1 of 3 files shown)


torkelrogstad (approved):
Thanks! Making some adjustments here.

PR #530, opened 2026-08-12
validator: fall back to JSON-RPC when REST is disabled

### LINKED ISSUE #212: validator: implement support for syncing block headers both with and without REST API
If REST isn't enabled we should fallback to fetching via JSON-RPC.

ekulkisnek:
## Problem

The enforcer currently exits when Bitcoin Core returns 404 for its REST endpoint. REST is optional in Core, so nodes configured with `-norest` cannot run the validator even though the same headers are available through JSON-RPC.

## Fix

Treat a REST 404 as an unavailable optional client and fall back to batched `getblockhash` plus `getblockheader` calls. Preserve the existing REST path when it is enabled.

Because Core executes batch entries independently, validate every fallback segment before storing it:

- require exact response counts;
- require the first hash to match the requested starting block;
- verify each reported hash and height;
- recompute every header hash and verify parent adjacency;
- recheck the segment endpoint against the active chain.

The existing public `Validator::new` API remains unchanged; the app uses a separate optional-REST constructor.

## Tests

- Seven focused unit cases cover a valid segment, short responses, stale starts, mismatched reported/computed hashes, wrong heights, and broken parent links.
- A `-norest` integration trial verifies REST is actually disabled, proves startup synchronization, mines another block, and waits for continued validator synchronization.
- `cargo +nightly fmt --all -- --check`
- `cargo clippy --all-targets -- --deny warnings`
- strict nightly `unqualified_local_imports` clippy pass
- `cargo test --tests` (143 tests)
- focused RPC validation tests (7 passed)
- focused `rest_disabled_header_sync` integration test

## Scope

This changes only header transport selection. Block download, validation, and the REST-enabled path are unchanged.

Closes #212

### DIFF
--- a/app/main.rs
+++ b/app/main.rs
@@ -1213,7 +1213,7 @@ async fn main() -> Result<()> {
 
     let ts = tokio::time::Instant::now();
     let mut retries = 0;
-    loop {
+    let mainchain_rest_client = loop {
         // A Bitcoin Core node with a large datadir takes tens of seconds to
         // open its REST interface
         const RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(1);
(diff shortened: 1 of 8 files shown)


torkelrogstad (approved):
Thank you. I made some adjustments here. Going to run this through some benchmarks before merging.

PR #533, opened 2026-08-12
messages: use the BIP300 M2 tag

ekulkisnek:
## Problem

The LayerTwo-Labs BIP300 specification assigns M2 the four-byte header `D6 E1 C5 BF`. The enforcer currently uses `D6 E1 C5 DF` instead. As a result it does not recognize standards-compliant M2 ACKs and emits a different wire message when building coinbases.

## Fix

Correct the M2 tag constant to `D6 E1 C5 BF` for both parsing and serialization.

## Tests

- Added a raw wire-form regression test that does not derive its bytes from the implementation constant.
- `cargo clippy --all-targets -- -D warnings`
- `cargo test --workspace` (134 library tests plus app tests)
- nightly rustfmt check and `git diff --check`

## Scope

M2 wire identification only; no ACK policy or activation changes.

### DIFF
--- a/lib/messages.rs
+++ b/lib/messages.rs
@@ -78,7 +78,7 @@ pub struct M2AckSidechain {
 }
 
 impl M2AckSidechain {
-    pub const TAG: [u8; 4] = [0xD6, 0xE1, 0xC5, 0xDF];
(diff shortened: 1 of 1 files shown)


torkelrogstad:
Thanks for flagging this. I believe the best thing to do here would be to change the spec, however.

PR #549, opened 2026-08-17
Handle block invalidation during initial sync

torkelrogstad:
Supersedes https://github.com/LayerTwo-Labs/bip300301_enforcer/pull/396. Thanks to the original author for flagging this issue.

### DIFF
--- a/lib/validator/cusf_enforcer.rs
+++ b/lib/validator/cusf_enforcer.rs
@@ -8,11 +8,10 @@ use std::{
 use async_broadcast::TrySendError;
 use bitcoin::{Block, BlockHash, Transaction, Txid, hashes::Hash as _};
 use cusf_enforcer_mempool::cusf_enforcer::{
-    ConnectBlockAction, CusfEnforcer, DisconnectBlockAction, TxAcceptAction,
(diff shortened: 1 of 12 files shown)


lib/validator/cusf_enforcer.rs
```diff
@@ -35,6 +34,11 @@ use crate::{
 #[repr(transparent)]
 pub struct SyncError(#[from] task::error::Sync);
 
+#[derive(Debug, Diagnostic, Error)]
+#[error(transparent)]
+#[repr(transparent)]
+pub struct InvalidBlockReason(Box<Splittable<task::error::ConnectBlock>>);
```
Ash-L2L: This looks wrong, we should not be invalidating blocks due to fatal errors.
The inner type should be `<task::error::ConnectBlock as Split>::Jfyi`

Ash-L2L (changes requested):

lib/validator/cusf_enforcer.rs
torkelrogstad: 🤦 sorry

Ash-L2L (approved):

PR #561, opened 2026-08-22
Expand sidechain voting policy, add bundle voting policies

torkelrogstad:
Block producers need more knobs for tuning precisely what to vote on.


Wired4ncer:
Block producer here, adding a data point on why these knobs matter before beta.

We run a pool on alphanet that has mined roughly half of the blocks since BIP-300 activation (measured 15,490 of 32,833 on 27 Aug). Withdrawal bundles are voted on by a simple majority of a 144-block window, so whatever our producer does on bundles is close to decisive on its own.

Today the enforcer gives us exactly two choices:

- `ack_all_proposals = true`: ACK every proposal, including ones that evict a live sidechain, and upvote the first pending bundle on every sidechain without looking at it.
- `ack_all_proposals = false`: choose proposals one by one, but the M4 is never emitted, so we abstain on every withdrawal everywhere.

There is no setting that is selective on proposals and still votes on bundles. On alphanet that means either every withdrawal passes because of us or none can, and beta on 20 Sep is when the first real bundles show up.

This PR is the shape we need. The policy we intend to publish for our miners is `NEW_SLOTS` for proposals (ACK new sidechains, hold evictions for an explicit decision) and `NONE` or `KNOWN` for bundles, with the per-bundle `SetWithdrawalBundleAck` used when miners direct an upvote. Every one of those already exists in this draft.

Happy to run it on our alphanet producer as soon as there is a rebased build, and to report back what the coinbase actually carries. If it helps sequencing, #593 is a smaller interim change to the same code path (upvote only bundles the node holds and can pay) that we could test in the meantime.

